PR #3, opened 2023-01-30
Add fake interface for workflow outputs, temporarily

Fryguy:


agrare:
This is no longer necessary now that the outputs from the steps are properly processed according to the ASL language and saved in the workflow instance record.

PR #6, opened 2023-03-14
Add specs for Workflow and WorkflowInstance

agrare:


agrare:
Cross-repo is green on these specs, merging

PR #7, opened 2023-03-15
Credentials references

agrare:
Allow users to add references to `Authentication` records in their `Workflow#credentials` property without having to embed the credentials directly in the record.

This uses a JSON Path style so for example: `{"username" => "$.service-now.userid", "password" => "$.service-now.password"}` would query for an authentication record with the name `service-now` and then retrieve the `userid` and `password` fields.

It is critical that the credential names be unique within the tenant, something which isn't guaranteed by the current `Authentication` model so I've added the validation to the `Workflows::AutomationManager::Authentication` subclass.

I'm also attaching the credential records to the seeded `Workflows::AutomationManager` here so they have something to relate to but that isn't strictly required for this to work.

TODO:
- [x] Enforce RBAC when querying the credentials (tenant+user)

Depends:
- [x] https://github.com/ManageIQ/manageiq-providers-workflows/pull/12
~~- [ ] Add miq_group_id to workflow_instances~~

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
@@ -1,4 +1,4 @@
-class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::AutomationManager::Authentication
+class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::EmbeddedAutomationManager::Authentication
   validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
             :format => {:with => /\A[\w\-]+\z/i, :message => N_("may contain only alphanumeric and _ - characters")}
 end
--- a/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
@@ -1,0 +1,3 @@
+class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::AutomationManager::Authentication
+  validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]}
+end
--- a/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
@@ -1,4 +1,4 @@
 class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::EmbeddedAutomationManager::Authentication
-  validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
+  validates :ems_ref, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
             :format => {:with => /\A[\w\-]+\z/i, :message => N_("may contain only alphanumeric and _ - characters")}
 end
--- a/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/authentication.rb
@@ -1,3 +1,4 @@
 class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::AutomationManager::Authentication
-  validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]}
+  validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
+            :format => {:with => /\A[\w\-]+\z/i, :message => N_("may contain only alphanumeric and _ - characters")}
 end
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
@@ -47,7 +47,7 @@ class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < Mana
 
         authentication.send(field)
       else
-        ManageIQ::Password.try_decrypt(val)
+        val
       end
     end
 
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
@@ -40,9 +40,9 @@ class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < Mana
   def run
     creds = credentials&.transform_values do |val|
       if val.start_with?("$.")
-        name, field = val.match(/^\$\.(?<name>.+)\.(?<field>.+)$/).named_captures.values_at("name", "field")
+        ems_ref, field = val.match(/^\$\.(?<ems_ref>.+)\.(?<field>.+)$/).named_captures.values_at("ems_ref", "field")
 
-        authentication = Rbac.filtered(manager.authentications.where(:name => name), :userid => run_by_userid, :miq_group_id => miq_group_id).first
+        authentication = Rbac.filtered(manager.authentications.where(:ems_ref => ems_ref), :userid => run_by_userid, :miq_group_id => miq_group_id).first
         raise ActiveRecord::RecordNotFound, "Couldn't find Authentication" if authentication.nil?
 
         authentication.send(field)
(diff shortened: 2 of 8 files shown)


app/models/manageiq/providers/workflows/automation_manager/authentication.rb
```diff
@@ -0,0 +1,3 @@
+class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::AutomationManager::Authentication
+  validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]}
```
agrare, line 2: TODO enforce a format that doesn't allow for spaces/periods so the JSON path regex will work, thinking non-fqdn hostname pattern would be good (something like `[a-zA-Z0-9\-]`)

Fryguy: :+1: This also will prevent json injection attacks

spec/models/manageiq/providers/workflows/automation_manager/workflow_instance_spec.rb
```diff
@@ -55,10 +51,44 @@
...
   describe "#run" do
-    it "test" do
+    it "sets the status to success" do
       workflow_instance.run
 
       expect(workflow_instance.reload.status).to eq("success")
     end
+
+    context "with a Credentials property in the workflow_content" do
+      let(:credentials) { {"username" => "$.my-credential.userid", "password" => "$.my-credential.password"} }
```
Fryguy, line 61: My only concern here is that it's not really discoverable at the moment, but I can't think of anything better.

agrare: There is a `GET /api/authentications` endpoint with `embedded_automation_manager_credentials_*` role permissions would that work?

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -39,7 +39,15 @@ def queue_callback(state, message, _result)
 
   def run
     creds = credentials&.transform_values do |val|
-      ManageIQ::Password.try_decrypt(val)
+      if val.start_with?("$.")
+        name, field = val.match(/^\$\.(?<name>.+)\.(?<field>.+)$/).named_captures.values_at("name", "field")
+
+        # TODO: this isn't filtering out authentication records from other tenants
+        authentication = Rbac.filtered_object(ext_management_system.authentications.find_by!(:name => name), :userid => userid, :miq_group => miq_group, :class => Authentication)
```
agrare, line 46: @kbrock I wonder if you could educate me on how `Rbac.filtered_object` works, this isn't filtering out auth records from other tenants

agrare: I _think_ this is because there is no `Authentication` in `TENANT_ACCESS_STRATEGY` so we're returning everything.  If I add `'Authentication' => :ancestor_ids` then this works but I don't know if that will break other things

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -39,7 +39,18 @@ def queue_callback(state, message, _result)
...
+        name, field = val.match(/^\$\.(?<name>.+)\.(?<field>.+)$/).named_captures.values_at("name", "field")
+
+        # TODO: this isn't filtering out authentication records from other tenants
+        authentication = ext_management_system.authentications.find_by!(:name => name)
+        authentication = Rbac.filtered_object(authentication, :userid => userid, :miq_group => miq_group, :class => Authentication)
+        raise ActiveRecord::RecordNotFound if authentication.nil?
+
+        authentication.send(field)
+      else
+        ManageIQ::Password.try_decrypt(val)
```
agrare, line 52: NOTE now that we allow external credentials do we want to allow inline encrypted credentials?

app/models/manageiq/providers/workflows/automation_manager/authentication.rb
```diff
@@ -0,0 +1,4 @@
+class ManageIQ::Providers::Workflows::AutomationManager::Authentication < ManageIQ::Providers::AutomationManager::Authentication
+  validates :name, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
+            :format => {:with => /\A[\w\-]+\z/i, :message => N_("may contain only alphanumeric and _ - characters")}
```
agrare, line 3: NOTE leave name as a user friendly description and use `#ems_ref` as the "unique" value for the purposes of querying from the workflow Credentials reference

app/models/manageiq/providers/workflows/automation_manager/authentication.rb
agrare, line 1: NOTE what will these authentication record be tied to, users might not have visibility to the `ManageIQ::Providers::Workflows::AutomationManager` instance but they have to be able to create authentication records

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
kbrock: I would prefer that we don't allow any passwords in requests/responses, and that we force this workflow.
But that is probably unrealistic.

app/models/manageiq/providers/workflows/automation_manager/authentication.rb
agrare: @Fryguy what if we associate these with a Tenant / MiqGroup ?

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: It seems like dropping the `:class => Authentication` removed the need for `TENANT_ACCESS_STRATEGY` :shrug:

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -39,7 +39,16 @@ def queue_callback(state, message, _result)
 
   def run
     creds = credentials&.transform_values do |val|
-      ManageIQ::Password.try_decrypt(val)
+      if val.start_with?("$.")
+        name, field = val.match(/^\$\.(?<name>.+)\.(?<field>.+)$/).named_captures.values_at("name", "field")
+
+        authentication = Rbac.filtered_object(ext_management_system.authentications.find_by!(:name => name), :userid => userid, :miq_group_id => miq_group_id)
```
agrare, line 45: NOTE I had to use `miq_group_id` here instead of `miq_group` because we dropped the `OwnershipMixin` from the core PR pending a discussion about how we want to RBAC these

kbrock: can we do a `where` instead of `find_by?`

Where will only bring back the objects once.
find_by will bring back all records, send them back to the db, and then download again.

agrare: Huh? `find_by` does a `SELECT LIMIT 1` not sure how that brings back all the records

agrare:
> Is there a way for us to create a foreign entity (hopefully one that has rbac already on it) and let the user use that? GenericProvider (new) or GenericObject comes to mind.

I think that greatly complicates the authoring of Workflows, since rather than `$.one-unique-field` they'll have to deal with multiple intermediate models each with different rules for uniqueness.  Uniquely referencing a VM vs a Host vs an EMS vs a Generic Object are all slightly different.

kbrock:
Yes, this does seem notably more complicated.

ok, so I'm thinking the authtype will dictate how the workflow connects to a host. Some use ssh keys, and others use username/password. Think that would be out of the hands of the workflow author and a `connect($authrecord)` would be used

We were talking about supporting an existing ems (which has multiple authentication types).
The secondary object (host, vm, generic) seems the only addition of complexity though. But it is taking complex and increasing it.

For the existing solution, the user would see what in the dropdown?
"#{ems.name} - #{object.name || ""} - #{auth_type}" maybe userid

Almost feel like `ems_id` needs to be in authentications. (yes, many times this will be the resource_id or indirectly associated with the resource_id)

agrare:
I think I'd rather create a new `Credentials` model if we aren't able to use Authentications due to Rbac/Tenancy.

So the question is, are we not able to use Authentications due to Rbac/Tenancy?

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: Okay done

agrare:
Now that we're using `ConfigurationScriptBase` there is an authentications join-table, I figure to get this in without all of the RBAC/OwnershipMixin stuff we could have whoever is managing the Workflow associate credential records with it "manually" for now.  Really just want to unblock this PR somehow.

agrare:
@Fryguy could you take a look at this one?

kbrock:
For global providers (automate, workflows), they are seeded and linked to the global tenant. we can only use Rbac directly on this object
For tenant based providers (vmware, openshift), do we want to use rbac from the ems?
Or is it possible copy it across from the provider to the auth record? So it is simpler/consistent?

Lets get this nailed down.
This has stalled us (Adam) too much

PR #10, opened 2023-03-23
Prepare initial third party services

Fryguy:
Prepare initial third party services


agrare:
This repo will fail until https://github.com/ManageIQ/manageiq/pull/22316 is merged, merging on red

PR #12, opened 2023-04-07
Use ConfigurationScripts for Workflows

agrare:
Depends:
- [x] https://github.com/ManageIQ/manageiq-schema/pull/688
- [x] https://github.com/ManageIQ/manageiq/pull/22477

https://github.com/ManageIQ/manageiq/issues/22311

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow.rb
@@ -4,7 +4,7 @@ class ManageIQ::Providers::Workflows::AutomationManager::Workflow < ManageIQ::Pr
     name = json["Comment"]
 
     workflows_automation_manager = ManageIQ::Providers::Workflows::AutomationManager.first
-    create!(:manager => workflows_automation_manager, :name => name, :payload => json, :payload_type => "json", **kwargs)
+    create!(:manager => workflows_automation_manager, :name => name, :payload => json.to_json, :payload_type => "json", **kwargs)
   end
 
   def execute(run_by_userid: "admin", inputs: {})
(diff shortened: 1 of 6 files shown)


app/models/manageiq/providers/workflows/automation_manager/workflow.rb
```diff
@@ -1,7 +1,15 @@
-class ManageIQ::Providers::Workflows::AutomationManager::Workflow < Workflow
+class ManageIQ::Providers::Workflows::AutomationManager::Workflow < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptPayload
```
agrare, line 1: NOTE the embedded_ansible provider calls these `Playbooks` [[ref]](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/playbook.rb) so kept with the theme and called these Workflows

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-schema/pull/688

app/models/manageiq/providers/workflows/automation_manager/workflow.rb
```diff
@@ -1,7 +1,15 @@
-class ManageIQ::Providers::Workflows::AutomationManager::Workflow < Workflow
-  def execute(userid: "admin", inputs: {})
+class ManageIQ::Providers::Workflows::AutomationManager::Workflow < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptPayload
+  def self.create_from_json!(json, **kwargs)
+    json = JSON.parse(json) if json.kind_of?(String)
+    name = json["Comment"]
```
Fryguy, line 4: Unrelated to this PR, but what happens if the Comment doesn't exists / is blank?

Fryguy (approved):

PR #15, opened 2023-04-26
Drop lib/gem-name.rb in favor of lib/gem/name.rb

jrafanie:
Adopt convention already supported by bundler and required by zeitwerk so we don't need to tell zeitwerk to ignore this lib/gem-name.rb file.

### DIFF
--- a/lib/manageiq/providers/workflows.rb
+++ b/lib/manageiq/providers/workflows.rb
@@ -1,3 +1,6 @@
+require "manageiq/providers/workflows/engine"
+require "manageiq/providers/workflows/version"
+
 module ManageIQ
   module Providers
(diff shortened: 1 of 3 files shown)


lib/manageiq/providers/workflows.rb
```diff
@@ -1,3 +1,6 @@
+require "manageiq/providers/workflows/engine"
+require "manageiq/providers/workflows/version"
```
kbrock, line 2: For engines, do we `require` our dependencies?

jrafanie: yes, I don't think you want to autoload/reload engine/version

PR #18, opened 2023-05-10
Rename `Workflow#execute` to `#run`

agrare:
Maintain compatibility with `EmbeddedAnsible::Playbook#run`


app/models/manageiq/providers/workflows/automation_manager/workflow.rb
```diff
@@ -7,7 +7,7 @@ def self.create_from_json!(json, **kwargs)
     create!(:manager => workflows_automation_manager, :name => name, :payload => json.to_json, :payload_type => "json", **kwargs)
   end
 
-  def execute(run_by_userid: "admin", inputs: {})
+  def run(inputs = {}, userid = "system")
```
agrare, line 10: NOTE I intended for this to be `system` originally, that was simply a bug

PR #20, opened 2023-05-22
Add support for syncing ASL workflows from git repo

agrare:
Add `#sync` support to Workflows `ConfigurationScriptSources`

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22524


app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -1,2 +1,36 @@
...
+    update!(:status => "running")
+
+    transaction do
+      current = configuration_script_payloads.index_by(&:name)
+
+      git_repository.update_repo
+      git_repository.with_worktree do |worktree|
+        worktree.ref = scm_branch
+        worktree.blob_list.each do |filename|
+          next if filename.start_with?(".") || !filename.end_with?(".asl")
```
Fryguy, line 18: So, something like `.example.asl` will be skipped?  Curious why the `.` is necessary to check?

agrare: Yes, this is to skip hidden files/folders.

For compatibility with EmbeddedAnsible: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/configuration_script_source.rb#L107 (they also have specs for this case)

spec/models/manageiq/providers/workflows/automation_manager/configuration_script_source_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+        :author     => {:email => "admin@localhost", :name => "admin"}
+      )
+    end
+
+    def dummy_workflow_data_for(filename)
+      <<~WORKFLOW_DATA
+        {"Comment": "hello world"}
+      WORKFLOW_DATA
+    end
+  end
```
agrare, line 582: NOTE lifted most of this from `Spec::Support::FakeAnsibleRepo` to get these specs working.  I don't think that is the correct place for that since the payload and file structure is essentially hard-coded in core there.  A base class or mixin which only handles the `Rugged` bits and allowed the caller to define the file structure and content would be more reusable.

Fryguy: Yeah that would be nice - FakeGitRepo vs FakeAnsibleRepo

agrare: @Fryguy opened https://github.com/ManageIQ/manageiq/pull/22524

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/22524

Fryguy:
@agrare Can you rebase?

PR #23, opened 2023-06-01
Invoke service before after callbacks

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-workflows/pull/22

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
@@ -1,6 +1,12 @@
 class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScript
-  def run_queue(zone: nil, role: "automation")
+  def run_queue(zone: nil, role: "automation", object: nil)
     raise _("run_queue is not enabled") unless Settings.prototype.ems_workflows.enabled
+
+    args = {:zone => zone, :role => role}
+    if object
+      args[:object_type] = object.class.name
+      args[:object_id]   = object.id
+    end
 
     queue_opts = {
       :class_name  => self.class.name,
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -1,14 +1,20 @@
...
+    end
+
     queue_opts = {
       :class_name  => self.class.name,
       :instance_id => id,
       :method_name => "run",
       :role        => role,
       :zone        => zone,
-      :args        => [{:zone => zone, :role => role}],
+      :args        => [args],
```
Fryguy, line 17: I don't understand the zone and role being in the args, when they're already in the queue item

Fryguy: Oh it's cause run calls run_queue ? (that's really backwards with respect to our method naming - usually run_queue calls run but on the queue)

Fryguy:
dependency PR was merged - so this can be rebased.

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: It is because we have to put next state on the queue, so run_queue does put run on the queue but run can "requeue"

Fryguy: ohhhh - ok - it's still weird but I get it now.

Fryguy:
Backported to `petrosian` in commit 1f3a3b1780dcc30523f50c9ac6eddb5fff46ffbf.

```text
commit 1f3a3b1780dcc30523f50c9ac6eddb5fff46ffbf
Date:   Wed Jun 7 14:24:43 2023 -0400

    Merge pull request #23 from agrare/invoke_service_before_after_callbacks
    
    Invoke service before after callbacks
    
    (cherry picked from commit 0f53abd56b9ba31bdc724a552410bcdfe0d8bf6d)
```

PR #25, opened 2023-06-07
Allow creation of Credentials for Workflows

agrare:
Creating an authentication requires defining `params_to_attributes` https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_automation_manager/authentication.rb#L28

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/credential.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/credential.rb
@@ -2,9 +2,70 @@ class ManageIQ::Providers::Workflows::AutomationManager::Credential < ManageIQ::
   validates :ems_ref, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
             :format => {:with => /\A[\w\-]+\z/i, :message => N_("may contain only alphanumeric and _ - characters")}
 
-  FRIENDLY_NAME = "Workflows Credential"
+  FRIENDLY_NAME = "Workflows Credential".freeze
+
+  COMMON_ATTRIBUTES = [
+    {
+      :component  => 'text-field',
+      :label      => N_('Name'),
+      :helperText => N_('Name of this credential'),
+      :name       => 'name',
+      :id         => 'name'
+    },
+    {
+      :component  => 'text-field',
+      :label      => N_('Reference'),
+      :helperText => N_('Unique reference for this credential'),
+      :name       => 'ems_ref',
+      :id         => 'ems_ref'
+    },
+    {
+      :component  => 'text-field',
+      :label      => N_('Username'),
+      :helperText => N_('Username for this credential'),
+      :name       => 'userid',
+      :id         => 'userid'
+    },
+    {
+      :component  => 'password-field',
+      :label      => N_('Password'),
+      :helperText => N_('Password for this credential'),
+      :name       => 'password',
+      :id         => 'password',
+      :type       => 'password'
+    },
+    {
+      :component      => 'password-field',
+      :label          => N_('Private key'),
+      :helperText     => N_('RSA or DSA private key to be used instead of password'),
+      :componentClass => 'textarea',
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/workflows/automation_manager/credential.rb
```diff
@@ -1,4 +1,72 @@
...
+      :label      => N_('Private key passphrase'),
+      :helperText => N_('Passphrase to unlock SSH private key if encrypted'),
+      :name       => 'auth_key_password',
+      :id         => 'auth_key_password',
+      :maxLength  => 1024,
+      :type       => 'password'
+    }
+  ].freeze
+
+  API_ATTRIBUTES = COMMON_ATTRIBUTES
```
agrare, line 57: NOTE we don't have any "EXTRA_ATTRIBUTES" so it seemed silly to define it then not use it.
The only reference I found to it other than `API_ATTRIBUTES = COMMON+EXTRA` was in AWX/Tower in the refresh process `inventory_object.options = inventory_object.type.constantize::EXTRA_ATTRIBUTES.keys.each_with_object({}) do |k, h|` so we don't seem to need it here.

app/models/manageiq/providers/workflows/automation_manager/credential.rb
```diff
@@ -1,4 +1,72 @@
...
+
+  API_OPTIONS = {
+    :label      => N_('Workflows'),
+    :type       => 'cloud',
```
agrare, line 61: TODO I'm not sure what this type is used for exactly yet, but I see `scm`, `vault`, and `cloud` where cloud includes some very non-cloudy things like VMware and RHV.

Fryguy: This is a field that came from AWX/Tower because they differentiate credential types.  If we can ignore it, we probably should.

app/models/manageiq/providers/workflows/automation_manager/credential.rb
```diff
@@ -1,4 +1,72 @@
...
+    :attributes => API_ATTRIBUTES
+  }.freeze
+
+  def self.params_to_attributes(params)
+    allowed_params     = API_ATTRIBUTES.pluck(:id)
```
Fryguy, line 66: WOW - I had no idea this existed for regular Arrays-of-Hashes.  Seems active_support offers this. 🆒

Fryguy (approved):

app/models/manageiq/providers/workflows/automation_manager/credential.rb
```diff
@@ -1,4 +1,72 @@
 class ManageIQ::Providers::Workflows::AutomationManager::Credential < ManageIQ::Providers::EmbeddedAutomationManager::Authentication
   validates :ems_ref, :presence => true, :uniqueness_when_changed => {:scope => [:tenant_id]},
             :format => {:with => /\A[\w\-]+\z/i, :message => N_("may contain only alphanumeric and _ - characters")}
+
+  FRIENDLY_NAME = "Workflows Credential"
```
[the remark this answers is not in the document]
agrare: Done

app/models/manageiq/providers/workflows/automation_manager/credential.rb
agrare: :+1: removed

Fryguy:
Backported to `petrosian` in commit 91f33a59eadc9231a94d64df898a50c2b2ec448b.

```text
commit 91f33a59eadc9231a94d64df898a50c2b2ec448b
Date:   Fri Jun 16 17:43:46 2023 -0400

    Merge pull request #25 from agrare/allow_creation_of_credentials_for_workflows
    
    Allow creation of Credentials for Workflows
    
    (cherry picked from commit e2545eec2191fe182cf0f1366b2979d349663390)
```

PR #26, opened 2023-06-15
Always allow seeding even without prototype flag

Fryguy:
This is in the background if the user hasn't set the prototype flag, so they won't realize it anyway. However, when they do enable the flag they can then hit the ground running without having to wait for a resed.

@agrare Please review.

### DIFF
--- a/lib/manageiq/providers/workflows.rb
+++ b/lib/manageiq/providers/workflows.rb
@@ -5,8 +5,6 @@ module ManageIQ
   module Providers
     module Workflows
       def self.seed
-        return unless Settings.prototype.ems_workflows.enabled
-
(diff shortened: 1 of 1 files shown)


agrare (approved):

Fryguy:
Backported to `petrosian` in commit 87fa44deb11114ccdb449e22958066a05672238a.

```text
commit 87fa44deb11114ccdb449e22958066a05672238a
Author: Adam Grare <adam@grare.com>
Date:   Thu Jun 15 14:57:52 2023 -0400

    Merge pull request #26 from Fryguy/always_seed
    
    Always allow seeding even without prototype flag
    
    (cherry picked from commit f1618887a1e3993bda330d4195bee34878dfe550)
```

PR #28, opened 2023-06-21
Set the Floe::Workflow::Runner based on platform

agrare:
Set the workflow runner based on the platform we're running on

Depends on:
- [x] https://github.com/ManageIQ/floe/pull/48
- [x] New release of Floe https://github.com/ManageIQ/floe/releases/tag/v0.2.0
- [x] https://github.com/ManageIQ/manageiq/pull/22603

### DIFF
--- a/config/initializers/floe_docker_runner.rb
+++ b/config/initializers/floe_docker_runner.rb
@@ -1,0 +1,2 @@
+require "floe"
+Floe::Workflow::Runner.docker_runner = ManageIQ::Providers::Workflows::Engine.floe_docker_runner
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -33,6 +33,26 @@ module ManageIQ
         def self.seedable_classes
           %w[ManageIQ::Providers::Workflows]
         end
+
+        def self.floe_docker_runner
+          require "miq_environment"
+          require "floe"
+
+          if MiqEnvironment::Command.is_podified?
+            host = ENV.fetch("KUBERNETES_SERVICE_HOST")
+            port = ENV.fetch("KUBERNETES_SERVICE_PORT")
+
+            Floe::Workflow::Runner::Kubernetes.new(
+              "server"     => URI::HTTPS.build(:host => host, :port => port).to_s,
+              "token_file" => "/run/secrets/kubernetes.io/serviceaccount/token",
+              "ca_cert"    => "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+            )
+          elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
+            Floe::Workflow::Runner::Podman.new
+          else
+            Floe::Workflow::Runner::Docker.new
+          end
+        end
       end
     end
   end


app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -66,6 +66,7 @@ def run(zone: nil, role: "automation", object_type: nil, object_id: nil)
       end
     end
 
+    Floe::Workflow::Runner.docker_runner = floe_docker_runner
```
agrare, line 69: NOTE I'm not sure the best place to set this, maybe an `Engine` initializer?

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -99,4 +102,21 @@ def run(zone: nil, role: "automation", object_type: nil, object_id: nil)
 
     run_queue(:zone => zone, :role => role, :object => object) if next_state.present?
   end
+
+  class << self
+    def floe_docker_runner
+      @floe_docker_runner ||= begin
```
Fryguy, line 108: Super-minor, but the begin is redundant, since the if condition is effectively a single expression.

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -99,4 +102,21 @@ def run(zone: nil, role: "automation", object_type: nil, object_id: nil)
...
+        if MiqEnvironment::Command.is_podified?
+          token_file = "/run/secrets/kubernetes.io/serviceaccount/token"
+          server     = URI::HTTPS.build(:host => ENV["KUBERNETES_SERVICE_HOST"], :port => ENV["KUBERNETES_SERVICE_PORT"])
+
+          Floe::Workflow::Runner::Kubernetes.new("token_file" => token_file, "server" => server)
+        elsif MiqEnvironment::Command.is_appliance?
+          Floe::Workflow::Runner::Podman.new
+        else
+          Floe::Workflow::Runner::Docker.new
+        end
```
Fryguy, line 118: @bdunne Since you run podman locally, I recall that we were trying to avoid this kind of logic, instead checking if podman is installed first, then falling back to docker.  Do you recall where / how we did that?  I think it was just a bunch of `which` statements

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -99,4 +102,21 @@ def run(zone: nil, role: "automation", object_type: nil, object_id: nil)
 
     run_queue(:zone => zone, :role => role, :object => object) if next_state.present?
   end
+
+  class << self
+    def floe_docker_runner
+      @floe_docker_runner ||= begin
+        if MiqEnvironment::Command.is_podified?
```
Fryguy, line 109: This is interesting with respect to the monolithic image.  In that case, we might do `docker` and require the monolithic image be run with the docker socket?  Not sure what to do there...perhaps @bdunne has thoughts.  Since that's not a primary use case, though we can just `TODO` it for now, and I won't let that hold up merge.

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
Fryguy: yeah it feels like this can be a one-time setup thing - not sure where it should live - perhaps a class method like `init_docker_runner`, which only runs once?  We do set up the Floe logger in an Engine initializer, so perhaps that's a good spot.

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: Actually that case should be handled, we have `is_container?` which covers the monolithic docker container and `is_podified?` checks that we are on k8s (it checks the `/run/secrets/kubernetes.io/serviceaccount/token` service account token)

agrare: I picked `is_podified?` specifically to be k8s and not monolithic docker

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: We could add podman, docker, and kubectl to `MiqEnvironment::Command.supports_*` that is where `supports_command?()` is

agrare:
@miq-bot cross-repo-tests including ManageIQ/floe@master

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-workflows/pull/28 including ManageIQ/floe

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-workflows/pull/28 including ManageIQ/floe

lib/manageiq/providers/workflows/engine.rb
```diff
@@ -24,6 +24,7 @@ def self.init_loggers
 
           require "floe"
           Floe.logger = $workflows_log
+          Floe::Workflow::Runner.docker_runner = floe_docker_runner
```
agrare, line 27: NOTE this obviously shouldn't live in `init_loggers` but until we have another method like maybe just `.init_plugin` this will do for testing.

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-workflows/pull/28 ManageIQ/floe

agrare:
Hm no idea why cross-repo isn't overriding the floe gem, will have to investigate that

agrare:
@miq-bot cross-repo-tests ManageIQ/floe@master

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
kbrock: I think alternatively you can put on the next line
```ruby
@floe_docker_runner ||=
  if MiqEnvironment::Command.is_podified?
    Floe::Workflow::Runner::Kubernetes.new
  elsif MiqEnvironment::Command.is_appliance?
    Floe::Workflow::Runner::Podman.new
  else
    Floe::Workflow::Runner::Docker.new
  end
```

or of course

@floe_docker_runner ||= new_runner_instance

def runner_class
  if MiqEnvironment::Command.is_podified?
    Floe::Workflow::Runner::Kubernetes.new
  elsif MiqEnvironment::Command.is_appliance?
    Floe::Workflow::Runner::Podman.new
  else
    Floe::Workflow::Runner::Docker.new
  end
```

or `runner_class.new` - which ever floats your boat.
probably nice to have the `new` hidden for future shenanigans. 

also
does it make sense to have
```ruby
class MiqEnvironment::Command
  def self.mode
    returns :kubernetes, :podman, :docker
  end
end
```

agrare: Yeah the current code looks very different from this older diff.  I do like having something like a `container_runner_command` in `MiqEnvironment`

lib/manageiq/providers/workflows/engine.rb
Fryguy: You can create an initializer the "normal" way via config/initializers.  Downside, is it would be run inside all instances.

I'd be ok with an init_plugin method, but I'm not sure how it would be different from a config/initializer.

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
Fryguy: Yeah that would be ok with me

agrare: https://github.com/ManageIQ/manageiq/pull/22603

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
Fryguy: I forgot when I said this that plugins support _normal_ config/initializers, but as I said in the other comment, it would run on any initialization - not sure if we can even delay load it though 🤷

lib/manageiq/providers/workflows/engine.rb
kbrock: interesting idea to set the values in the `config/initializers` (at install time)
rather than detecting the values.

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: Yeah we're calling all of the `Vmdb::Plugins.init_*` on every initialization anyway, I'll move it to `config/initializers` just wish there was a better place to put this that wasn't a.) on every init or b.) all the way down in WorkflowInstance#execute

Fryguy: Agreed - we need some sort of way to delay load things but run a one-time initialization when that thing is loaded.  🤔

config/initializers/floe_docker_runner.rb
```diff
@@ -0,0 +1,21 @@
+def floe_docker_runner
```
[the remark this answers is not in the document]
agrare: So "docker" in this case is talking about the resource that is being run so a docker image specified by "docker://..." in the ASL file, not about what ends up being run.  A little bit of name overloading unfortunately.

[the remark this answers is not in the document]
bdunne: I think the same still applies and would argue that `container://...` seems more appropriate.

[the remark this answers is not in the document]
agrare: If we did decide to change the resource scheme name for a docker image I'm not sure I'd start with this method.  We'd need to start with what we use in the ASL files first which would live in the floe gem so that'd be a place to start the discussion.

Podman does use `docker://` as the scheme for a remote image URI which is what we're using so this is at least a cross-vendor recognized format [[ref]](https://docs.podman.io/en/latest/markdown/podman-run.1.html)

> docker://docker-reference (Default) An image reference stored in a remote container image registry.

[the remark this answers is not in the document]
Fryguy: While I'd agree container:// is probably clearer, it is common practice to use docker:// (podman example above and also skopeo uses that format), so I agree with keeping docker://

Fryguy, line 1: @agrare I have a separate problem with this in that it is introducing a method at the top-level scope.  Can we move this method into a module/namespace (not sure if you can introduce `ManageIQ::Providers::Workflows` namespace here)?  Or perhaps move it as an engine method but keep it called here, such as `ManageIQ::Providers::Workflows::Engine.floe_docker_runner`?

Apologies in advance because I know my suggestions have been having you copy paste this all over creation :(

agrare: I moved this method to the `Workflows::Engine` class

config/initializers/floe_docker_runner.rb
```diff
@@ -0,0 +1,21 @@
...
+  require "miq_environment"
+
+  if MiqEnvironment::Command.is_podified?
+    host = ENV.fetch("KUBERNETES_SERVICE_HOST")
+    port = ENV.fetch("KUBERNETES_SERVICE_PORT")
```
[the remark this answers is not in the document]
agrare: `is_podified?` means kubernetes, `is_container?` means docker-in-general

https://github.com/ManageIQ/manageiq/blob/master/lib/miq_environment.rb#L58-L69

Fryguy (approved):

lib/manageiq/providers/workflows/engine.rb
```diff
@@ -33,6 +33,25 @@ def self.apply_logger_config(config)
         def self.seedable_classes
           %w[ManageIQ::Providers::Workflows]
         end
+
+        def self.floe_docker_runner
+          require "miq_environment"
```
Fryguy, line 38: Super-duper minor, and I won't hold up merge, but I think technically this needs a `require "floe"`, since the method uses it.  I know it's called in the initializer first anyway, so not a big deal.

agrare: Added

Fryguy:
Backported to `petrosian` in commit d371ad544f30c69055e6971d946c1f7e0367fdcb.

```text
commit d371ad544f30c69055e6971d946c1f7e0367fdcb
Date:   Mon Jul 24 15:02:55 2023 -0400

    Merge pull request #28 from agrare/set_floe_workflow_runner
    
    Set the Floe::Workflow::Runner based on platform
    
    (cherry picked from commit 2fc3cee077b346dfabb173ccc86301022b2a81e3)
```

PR #29, opened 2023-06-21
Fix creating Credentials from API

agrare:


app/models/manageiq/providers/workflows/automation_manager/credential.rb
```diff
@@ -62,8 +62,8 @@ class ManageIQ::Providers::Workflows::AutomationManager::Credential < ManageIQ::
   }.freeze
 
   def self.params_to_attributes(params)
-    allowed_params     = API_ATTRIBUTES.pluck(:id)
-    unpermitted_params = params.keys - allowed_params
+    allowed_params     = API_ATTRIBUTES.pluck(:id) + ["type"]
```
agrare, line 65: There is also a "type" holding the class name

agrare: No problem validating this, but is that something we do elsewhere?  I don't recall seeing that sort of check on other API payloads.  We wouldn't be in this class if the type wasn't already a subclass of this one

agrare: Actually no, there's no way we could get here if `"type"` wasn't already a valid subclass of this one https://github.com/ManageIQ/manageiq-api/blob/master/lib/services/api/authentication_service.rb#L4-L6

kbrock:
fixed by https://github.com/ManageIQ/manageiq/pull/22570

PR #5, opened 2023-03-06
Move workflow executing into manageiq-floe

agrare:
Depends on:
- [x] https://github.com/ManageIQ/floe/pull/33
- [x] https://github.com/ManageIQ/floe/pull/44

PR #35, opened 2023-07-24
Fix WorkflowInstance#output not being set

agrare:
After moving the state run logic into floe we no longer were setting the state output.  This was causing issues when the caller needs the output to populate e.g. a dynamic dialog dropdown.

https://github.com/ManageIQ/manageiq-providers-workflows/pull/5/files#diff-80ee5262679b29c698ee75771be7948fbfe8fcea9ff7c05ee83aaa9273d47885L89


Fryguy:
Backported to `petrosian` in commit 66bd179909799766133541e324b262d428bbfdd4.

```text
commit 66bd179909799766133541e324b262d428bbfdd4
Date:   Mon Jul 24 12:26:39 2023 -0400

    Merge pull request #35 from agrare/fix_missing_output
    
    Fix WorkflowInstance#output not being set
    
    (cherry picked from commit d4035e61c08f61fb99866606a16a958904ae90ae)
```

PR #39, opened 2023-07-30
Pass manageiq_api url in global context

agrare:
Set the manageiq_api_url in the global context so that it can be referenced by workflows.

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow.rb
@@ -12,6 +12,7 @@ class ManageIQ::Providers::Workflows::AutomationManager::Workflow < ManageIQ::Pr
 
     require "floe"
     context = Floe::Workflow::Context.new(:input => inputs)
+    context.execution["_manageiq_api_url"] = MiqRegion.my_region.remote_ws_url
 
     miq_task = instance = nil
     transaction do
--- a/spec/models/manageiq/providers/workflows/automation_manager/workflow_spec.rb
+++ b/spec/models/manageiq/providers/workflows/automation_manager/workflow_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe ManageIQ::Providers::Workflows::AutomationManager::Workflow do
         :type        => "ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance",
         :payload     => workflow.payload,
         :credentials => workflow.credentials,
-        :context     => {"Execution" => {"Input" => {}}, "State" => {}, "StateMachine" => {}, "StateHistory" => [], "Task" => {}},
+        :context     => {"Execution" => hash_including("Input" => inputs), "State" => {}, "StateMachine" => {}, "StateHistory" => [], "Task" => {}},
         :output      => {},
         :status      => "pending"
       )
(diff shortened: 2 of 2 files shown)


kbrock:
Is the thought that this will lineup with how the workflows access the api?

e.g.: https://github.com/ManageIQ/workflows-examples/blob/master/provision-vm-service/provision-vm.asl#L10-L11

agrare:
The idea here is that the user shouldn't have to manually specify the API_URL in their workflows (example usage https://github.com/ManageIQ/workflows-examples/blob/master/provision-vm-service/list-providers/list-providers.rb#L10)

NOTE these workflows-examples currently work because of the default to `localhost:3000` which is actually not possible outside of dev.

Fryguy:
I assume this is similar to how we pass these to regular automate methods?

One question here.  Will the token persist into the StatesHistory and can that be problematic?

Fryguy:
Note that I said that I wonder if we should pass the StatesHistory though the generic password masker just in case. 🤔

agrare:
> I assume this is similar to how we pass these to regular automate methods?

I started out merging these with the normal input, but based on how step-functions does input/output if they didn't include this in the OutputPath it wouldn't flow through to the rest of the states after the first so I decided it would be better to add the URL to the Context and it would be accessible with the double `$$.`

> One question here. Will the token persist into the StatesHistory and can that be problematic?

Not StatesHistory but it would be in the context object, I should merge this with credentials instead thanks for pointing that out.

agrare:
Since we already have a mechanism for mapping credentials I'm going to take care of the api_token in a follow-up PR

Here's an example of a workflow using the api_url from the context: https://github.com/ManageIQ/workflows-examples/compare/master...agrare:workflows-examples:pass_api_url_context#diff-fcf2b10d933d60255cfe87aa3603defddd838acf4b4035835d6f06d5945f52f8R14
```
      "Parameters": {
        "API_URL.$": "$$.Execution._manageiq_api_url",
```

And running on a non-default binding ip: `rails s -b 10.2.4.100`
```
  context:
   {"Task"=>{},
    "State"=>
     {"Guid"=>"9c74b5b5-37b7-45b6-9c7b-36b0a3d00c54",
      "Name"=>"ListProviders",
      "Input"=>{},
      "Output"=>{"values"=>{"3"=>"vcsim"}, "API_URL"=>"http://10.2.4.100:3000", "VERIFY_SSL"=>"[FILTERED]", "PROVIDER_TYPE"=>"ManageIQ::Providers::Vmware::InfraManager"},
      "Duration"=>4.957739790988853,
      "EnteredTime"=>"2023-09-26T18:39:13.053Z",
      "FinishedTime"=>"2023-09-26T18:39:18.010Z"},
    "Execution"=>{"Input"=>{}, "StartTime"=>"2023-09-26T18:39:13.052Z", "_manageiq_api_url"=>"http://10.2.4.100:3000"}
```

Fryguy:
Backported to `quinteros` in commit b220293d500c01a9340b7f32c46dc790292a1783.

```text
commit b220293d500c01a9340b7f32c46dc790292a1783
Date:   Thu Sep 28 09:47:12 2023 -0400

    Merge pull request #39 from agrare/pass_manageiq_api_url_in_parameters
    
    Pass manageiq_api url in global context
    
    (cherry picked from commit a7c24c7cdd3276be164508e0bbc07b7d45cc50ff)
```

PR #44, opened 2023-08-14
[WIP] Add a WorkflowDispatcher worker

agrare:
Add a dedicated worker to run embedded workflows

Depends on:
- [ ] https://github.com/ManageIQ/manageiq/pull/22656

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_dispatcher.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_dispatcher.rb
@@ -1,0 +1,16 @@
+class ManageIQ::Providers::Workflows::AutomationManager::WorkflowDispatcher < MiqQueueWorkerBase
+  include MiqWorker::ReplicaPerWorker
+
+  require_nested :Runner
+
+  self.required_roles     = %w[embedded_workflows]
+  self.default_queue_name = "workflows"
+
+  def self.kill_priority
+    MiqWorkerType::KILL_PRIORITY_GENERIC_WORKERS
+  end
+
+  def self.settings_name
+    :workflows_dispatcher
+  end
+end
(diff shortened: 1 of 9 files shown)


kbrock:
Goal of this is to setup a worker with privileges (security context) to create pods.

Fryguy:
I was hoping we could avoid creating a new worker, but perhaps it's inevitable.  Main reason is that we would need _another_ kubernetes service account, and we always tried to avoid that leaving that to only the orchestrator and the operator.  Let's discuss before too much time gets put into this solution.  cc @bdunne 

(FWIW, I eventually want the _orchestrator_ to get rid of the service account too and we'd only have the operator, but that might be a pipe dream).

app/models/manageiq/providers/workflows/automation_manager/workflow_dispatcher.rb
```diff
@@ -0,0 +1,16 @@
+class ManageIQ::Providers::Workflows::AutomationManager::WorkflowDispatcher < MiqQueueWorkerBase
+  include MiqWorker::ReplicaPerWorker
```
Fryguy, line 2: If we need a new worker, I'd prefer it were a single worker that doesn't have replicas, and is just one per region (which might be sort of automatic since we only have one EMS).  (I can't remember how `ReplicaPerWorker` acts compared to the other types)

This way we can keep the service account tied to a single, well defined, deployment/pod.

agrare: `ReplicaPerWorker` just means it isn't a per-ems worker.  If there is only a single allowed per scope is separate

agrare:
We decided to go a different direction with this and make this a more generic automate runner

PR #46, opened 2023-08-23
Override the podman image storage root directory

agrare:
By default non-root images are stored in the home directory of the non-root user.  On appliances as the manageiq user this is a very small partition and not many images can be stored here.

Depends on:
- [x] https://github.com/ManageIQ/floe/pull/90
- [x] https://github.com/ManageIQ/manageiq-rpm_build/pull/415

https://github.com/ManageIQ/manageiq/issues/22640

### DIFF
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -49,7 +49,12 @@ module ManageIQ
             )
           elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
             options = {}
-            options["network"] = "host" unless Rails.env.production?
+            if Rails.env.production?
+              options["root"] = Rails.root.join("data/containers/storage").to_s
+            else
+              options["network"] = "host"
+            end
+
             Floe::Workflow::Runner::Podman.new(options)
           else
             options = {}


lib/manageiq/providers/workflows/engine.rb
```diff
@@ -49,7 +49,13 @@ def self.floe_docker_runner
             )
           elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
             options = {}
-            options["network"] = "host" unless Rails.env.production?
+            if Rails.env.production?
+              options["root"]        = Rails.root.join("data/containers/storage").to_s
+              options["storage-opt"] = "ignore_chown_errors=true" # Fix issues with pulling images as a non-root user
```
agrare, line 54: NOTE I want to investigate this one, https://www.redhat.com/sysadmin/rootless-podman mentions adding an option to `/etc/subuid` and `/etc/subgid` that is the better way of doing this

agrare: https://github.com/ManageIQ/manageiq-rpm_build/pull/415

Fryguy:
Backported to `quinteros` in commit a1ce8bd3340674ab27f580892321c252c64a38b9.

```text
commit a1ce8bd3340674ab27f580892321c252c64a38b9
Date:   Tue Aug 29 12:20:15 2023 -0400

    Merge pull request #46 from agrare/podman_containers_storage_location
    
    Override the podman image storage root directory
    
    (cherry picked from commit 047c9b19bd0d2cb753cc8a1ed3083537577c807c)
```

PR #49, opened 2023-09-21
Strip trailing .$ from Credentials

agrare:
When transforming a payload_template style hash we have to strip off the
trailing .$ after interpolating the value.

PR #50, opened 2023-09-21
[WIP] Run workflows asyncrhonously on MiqAutomationWorker

agrare:
Depends on:
- [ ] https://github.com/ManageIQ/manageiq/pull/22710


app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -55,21 +57,21 @@ def run(args = {})
     object = object_type.constantize.find_by(:id => object_id) if object_type && object_id
     object.before_ae_starts({}) if object.present? && object.respond_to?(:before_ae_starts)
 
-    creds = credentials&.transform_values do |val|
```
kbrock, line 58: will this be flat or do we need to be recursive?

Not sure if I like hashes or $ better. Feels like we are mixing paradigms. The `.$` will use path and stuff, vs with our hash, we are going the other way.
(I'd prefer just going .$ and $, but I can see why we are going the hash route)

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -55,21 +57,21 @@ def run(args = {})
...
+        [key.chomp(".$"), authentication.send(credential_field)]
       else
-        val
+        [key, val]
       end
     end
 
     wf = Floe::Workflow.new(payload, context, creds)
-    wf.step
+    wf.run_nonblock
```
kbrock, line 74: 🎉

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -87,6 +89,6 @@ def run(args = {})
       object.after_ae_delivery(ae_result)
     end
 
-    run_queue(:zone => zone, :role => role, :object => object) unless wf.end?
+    run_queue(:zone => zone, :role => role, :object => object, :deliver_on => 1.minute.from_now.utc) unless wf.end?
```
kbrock, line 92: think I want to expose `wf.waiting`

spec/models/manageiq/providers/workflows/automation_manager/workflow_instance_spec.rb
```diff
@@ -134,8 +135,8 @@
             "FirstState" => {
               "Type"        => "Succeed",
               "Credentials" => {
-                "username" => "$.username",
-                "password" => "$.password"
+                "username.$" => "$.username",
+                "password.$" => "$.password"
```
kbrock, line 139: yea. this is beautiful.

wish the credentials was a little more similar

agrare: This was actually from a separate PR that I based this one on https://github.com/ManageIQ/manageiq-providers-workflows/pull/49

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: Yeah I was hoping to contain that knowledge inside the `.wait` method

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -87,6 +90,6 @@ def run(args = {})
       object.after_ae_delivery(ae_result)
     end
 
-    run_queue(:zone => zone, :role => role, :object => object) unless wf.end?
+    run_queue(:zone => zone, :role => role, :object => object, :deliver_on => 10.seconds.from_now.utc, :server_guid => MiqServer.my_server.guid) unless wf.end?
```
Fryguy, line 93: @agrare Is this what you mentioned to me?

So my question here is why we need the pinning and deliver time at all?

agrare: Because this re-queue while the docker/podman/k8s task is still running so we don't want it to spin (hence deliver on) and we need to run it on the same server so `podman inspect` will be able to see the running container.

Fryguy: ohhh I see - ok, this is what I wasn't understanding in our discussion yesterday because I thought it worked differently.

Fryguy (approved):

Fryguy:
Backported to `quinteros` in commit 0d1e031d66c96dcd4e55dac39782d73bbd989926.

```text
commit 0d1e031d66c96dcd4e55dac39782d73bbd989926
Date:   Wed Sep 27 11:16:50 2023 -0400

    Merge pull request #50 from agrare/run_workflows_on_miq_automation_worker
    
    Run workflows asyncrhonously on MiqAutomationWorker
    
    (cherry picked from commit be6f4d1e202f6ea61510a3488281ba62452e5f25)
```

PR #55, opened 2023-10-12
Pass the workflow service account

Fryguy:
@agrare @bdunne @kbrock Please review.

This PR hardcodes the value for now, but it should probably be passed down as an ENV var.  Not sure if we want to do this now or later.  @bdunne Thoughts?

Depends on 
- [x] https://github.com/ManageIQ/floe/pull/131 (and subsequent release)

Must be merged with
- [ ] https://github.com/ManageIQ/manageiq/pull/22740

### DIFF
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -42,12 +42,15 @@ module ManageIQ
             host = ENV.fetch("KUBERNETES_SERVICE_HOST")
             port = ENV.fetch("KUBERNETES_SERVICE_PORT")
 
-            Floe::Workflow::Runner::Kubernetes.new(
-              "server"     => URI::HTTPS.build(:host => host, :port => port).to_s,
-              "token_file" => "/run/secrets/kubernetes.io/serviceaccount/token",
-              "ca_cert"    => "/run/secrets/kubernetes.io/serviceaccount/ca.crt",
-              "namespace"  => File.read("/run/secrets/kubernetes.io/serviceaccount/namespace")
-            )
+            options = {
+              "server"               => URI::HTTPS.build(:host => host, :port => port).to_s,
+              "token_file"           => "/run/secrets/kubernetes.io/serviceaccount/token",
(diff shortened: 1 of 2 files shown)


lib/manageiq/providers/workflows/engine.rb
```diff
@@ -43,10 +43,11 @@ def self.floe_docker_runner
...
             Floe::Workflow::Runner::Kubernetes.new(
-              "server"     => URI::HTTPS.build(:host => host, :port => port).to_s,
-              "token_file" => "/run/secrets/kubernetes.io/serviceaccount/token",
-              "ca_cert"    => "/run/secrets/kubernetes.io/serviceaccount/ca.crt",
-              "namespace"  => File.read("/run/secrets/kubernetes.io/serviceaccount/namespace")
+              "server"                   => URI::HTTPS.build(:host => host, :port => port).to_s,
+              "token_file"               => "/run/secrets/kubernetes.io/serviceaccount/token",
+              "ca_cert"                  => "/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+              "namespace"                => File.read("/run/secrets/kubernetes.io/serviceaccount/namespace"),
+              "workflow_service_account" => "manageiq-default"
```
bdunne, line 50: This value can be productized using the app name in the CR, so it may not always be "manageiq"

Fryguy: Moved to an ENV var AUTOMATION_JOB_SERVICE_ACCOUNT

Fryguy:
Backported to `quinteros` in commit 81d76fda2a9b91e2f2f4fa9557c2fdae3962b5b3.

```text
commit 81d76fda2a9b91e2f2f4fa9557c2fdae3962b5b3
Date:   Thu Oct 12 15:40:56 2023 -0400

    Merge pull request #55 from Fryguy/add_workflow_service_account
    
    Pass the workflow service account
    
    (cherry picked from commit d73c48cb24e133c58879b572eab755882ff157f1)
```

PR #56, opened 2023-10-14
Fix issue where new service account ENV var is forced in every process

Fryguy:
ManageIQ::Providers::Workflows::Engine.floe_docker_runner is called in a config/initializer, which means that it is being executed in every Rails environment on boot. This includes things like `bin/rake evm:deployment_status`, which happens very early on in the orchestrator boot process.

When the orchestrator is on Kubernetes, it starts creating a Floe::Kubernetes::Runner, but then requires the new AUTOMATION_JOB_SERVICE_ACCOUNT, which doesn't exist as it is only meant to exist in the new AutomationWorker. Because of this the orchestrator fails to boot.

This commit allows that value to be nil, so it can move on. However, we should find a way to avoid creating the runner on every boot, which is a better solution.

### DIFF
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -47,7 +47,7 @@ module ManageIQ
               "token_file"           => "/run/secrets/kubernetes.io/serviceaccount/token",
               "ca_cert"              => "/run/secrets/kubernetes.io/serviceaccount/ca.crt",
               "namespace"            => File.read("/run/secrets/kubernetes.io/serviceaccount/namespace"),
-              "task_service_account" => ENV.fetch("AUTOMATION_JOB_SERVICE_ACCOUNT")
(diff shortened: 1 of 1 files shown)


Fryguy:
cc @bdunne @agrare - This happened after the merge of #55 

Moving forward we need to find a way to avoid creating the runner until we need it.  If we have to do it in an initializer, one fix might be to pass a proc to the docker_runner, and the first invocation can resolve it (possibly with a Mutex).

Fryguy:
Self-merging so we can backport and get a clean build.

Fryguy:
Backported to `quinteros` in commit 3be6f7fed16e6a2fdf363e9765ea94e990ac62de.

```text
commit 3be6f7fed16e6a2fdf363e9765ea94e990ac62de
Date:   Fri Oct 13 22:14:30 2023 -0400

    Merge pull request #56 from Fryguy/fix_service_account
    
    Fix issue where new service account ENV var is forced in every process
    
    (cherry picked from commit 2d88b90b41431f495c7af07ab8b9a44d37840142)
```

PR #57, opened 2023-10-27
Seed workflows from plugins' content/workflows

agrare:
Seed workflows from plugins' `content/workflows/**/*.asl` 

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22759

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -14,6 +14,8 @@ class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSour
   end
 
   def self.seed
+    return unless Settings.prototype.ems_workflows.seed_builtin_workflows
+
     manager = ManageIQ::Providers::Workflows::AutomationManager.in_my_region.first
     return if manager.nil?
 
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -1,6 +1,6 @@
 class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSource < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptSource
   FRIENDLY_NAME     = "Embedded Workflows Repository".freeze
-  BUILTIN_REPO_NAME = "ManageIQ".freeze
+  BUILTIN_REPO_NAME = "ManageIQ Built-in Workflows".freeze
 
   supports :update do
     _("Cannot update the built-in repository") if name == BUILTIN_REPO_NAME
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -2,6 +2,13 @@ class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSour
   FRIENDLY_NAME     = "Embedded Workflows Repository".freeze
   BUILTIN_REPO_NAME = "ManageIQ".freeze
 
+  supports :update do
+    _("Cannot update the built-in repository") if name == BUILTIN_REPO_NAME
+  end
+  supports :delete do
+    _("Cannot delete the built-in repository") if name == BUILTIN_REPO_NAME
+  end
+
   def self.display_name(number = 1)
     n_('Repository (Embedded Workflows)', 'Repositories (Embedded Workflows)', number)
   end
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -1,8 +1,16 @@
 class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSource < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptSource
-  FRIENDLY_NAME = "Embedded Workflows Repository".freeze
+  FRIENDLY_NAME     = "Embedded Workflows Repository".freeze
+  BUILTIN_REPO_NAME = "ManageIQ".freeze
 
   def self.display_name(number = 1)
     n_('Repository (Embedded Workflows)', 'Repositories (Embedded Workflows)', number)
+  end
+
+  def self.seed
+    manager = ManageIQ::Providers::Workflows::AutomationManager.in_my_region.first
+    return if manager.nil?
+
+    manager.configuration_script_sources.find_or_create_by!(:type => name, :name => BUILTIN_REPO_NAME)
   end
 
   def sync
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -5,6 +5,18 @@ def self.display_name(number = 1)
     n_('Repository (Embedded Workflows)', 'Repositories (Embedded Workflows)', number)
   end
 
+  def self.seed
+    manager = ManageIQ::Providers::Workflows::AutomationManager.in_my_region.first
+    return if manager.nil?
+
+    manager.configuration_script_sources.find_or_create_by!(:type => name, :name => "Embedded Workflows")
```
Fryguy, line 12: Not sure about the name here.  I think I'd prefer `ManageIQ` as the name similar to how we named the built-in automate domain as `ManageIQ`, but I'm not sure.  Productization-wise it's a little weird.

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -5,6 +5,18 @@ def self.display_name(number = 1)
...
+  end
+
+  def git_repository
+    return super unless name == "Embedded Workflows"
```
Fryguy, line 16: Let's move the "internal" name into a constant so it's shared with the create side.

app/models/manageiq/providers/workflows/automation_manager/workflow.rb
```diff
@@ -1,4 +1,26 @@
...
+  def self.seed
+    manager = ManageIQ::Providers::Workflows::AutomationManager.in_my_region.first
+    return if manager.nil?
+
+    repo = manager.configuration_script_sources.find_by(:name => "Embedded Workflows")
+    return if repo.nil?
+
+    to_delete = repo.configuration_script_payloads.pluck(:id)
+    Vmdb::Plugins.embedded_workflows_content.each do |filename|
+      relative_filename = filename.split("content/workflows/").last
```
Fryguy, line 11: What happens if 2 plugins provide the same workflow name?  Wondering if we should straight disallow that through specs, or perhaps we prefix the name with the plugin name?

Fryguy: Super minor, but Pathname does this for you...

```suggestion
      relative_filename = Pathname.new(filename).relative_path_from("content/workflows")
```

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
agrare: Sounds good, how to we productize the builtin automate domain?  We could use a Setting and then override that

app/models/manageiq/providers/workflows/automation_manager/workflow.rb
agrare: I updated the core method to yield Pathnames but I don't think `relative_path_from` works like this at least not for absolute paths
```
(byebug) filename.relative_path_from("content/workflows")
*** ArgumentError Exception: different prefix: "/" and "content/workflows"

nil
(byebug) filename
#<Pathname:/home/grare/adam/src/manageiq/manageiq-providers-workflows/content/workflows/hello_world/workflow.asl>
```

app/models/manageiq/providers/workflows/automation_manager/workflow.rb
```diff
@@ -1,4 +1,28 @@
...
+    return if repo.nil?
+
+    to_delete = repo.configuration_script_payloads.pluck(:id)
+    Vmdb::Plugins.embedded_workflows_content.each do |filename|
+      plugin, path = filename.to_s
+                             .match(/.+\/(?<plugin>.+)\/content\/workflows\/(?<path>.+)/)
+                             .values_at("plugin", "path")
+
+      payload  = File.read(filename)
+      workflow = repo.configuration_script_payloads.find_or_initialize_by(:name => "#{plugin}/#{path}")
```
agrare, line 16: NOTE we use `:name` as the unique attribute and set it to the file path to be consistent with how embedded ansible was behaving.

Now I'm wondering if we'd be better off setting `manager_ref` to the filename, treating that as unique, and setting `name` to the `"Comment"` value like we do in create_from_json below.  Would be a simple db migration to set `manager_ref = name` for all "embedded" configuration_scripts

agrare: This can be a follow-up not directly related to this

agrare:
Hey @Fryguy this has been out there for a while, what do you think?

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -1,10 +1,18 @@
...
 
   def self.display_name(number = 1)
     n_('Repository (Embedded Workflows)', 'Repositories (Embedded Workflows)', number)
   end
 
+  def self.seed
+    manager = ManageIQ::Providers::Workflows::AutomationManager.in_my_region.first
+    return if manager.nil?
+
+    manager.configuration_script_sources.find_or_create_by!(:type => name, :name => BUILTIN_REPO_NAME)
```
agrare, line 13: TODO add system to prevent user modification via UI/API

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -1,10 +1,18 @@
 class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSource < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptSource
-  FRIENDLY_NAME = "Embedded Workflows Repository".freeze
+  FRIENDLY_NAME     = "Embedded Workflows Repository".freeze
+  BUILTIN_REPO_NAME = "ManageIQ".freeze
```
agrare, line 3:
```suggestion
  BUILTIN_REPO_NAME = "ManageIQ Built-in Workflows".freeze
```

or
```suggestion
  BUILTIN_REPO_NAME = "ManageIQ Workflows".freeze
```

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -1,10 +1,18 @@
...
 
   def self.display_name(number = 1)
     n_('Repository (Embedded Workflows)', 'Repositories (Embedded Workflows)', number)
   end
 
+  def self.seed
```
agrare, line 9: TODO have "seed -> ConfigurationScriptSource#sync for built-in repo"

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -1,30 +1,42 @@
...
+    return unless Settings.prototype.ems_workflows.seed_builtin_workflows
+
+    manager = ManageIQ::Providers::Workflows::AutomationManager.in_my_region.first
+    return if manager.nil?
+
+    manager.configuration_script_sources
```
Fryguy, line 22: Why the .tap?  I assume you used it to return the seeded record but I thought we didn't care about the return value?

Fryguy (approved):
Just one small q

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
agrare: Yeah that's true, it was returning the seeded repository previously but we don't need to

agrare: Done

Fryguy:
This PR has been effectively backported to `radjabov` via the merge of master into radjabov.

PR #59, opened 2023-11-08
Check if a workflow payload is valid during sync

agrare:
When syncing a repository and creating workflows check if the payload is valid and if it isn't set a (hopefully) helpful error message.

Depends on:
- [x] https://github.com/ManageIQ/manageiq-schema/pull/710
~~https://github.com/ManageIQ/manageiq-schema/pull/712~~

TODO:
* We don't currently check if a configuration_script_payload "supports" running, we could enable that and check for validity there and use payload_error as the unsupported reason

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -85,18 +85,18 @@ class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSour
         [nil, err.message]
       end
 
-    return if payload_error
-
     description = floe_workflow&.comment
 
     configuration_script_payloads.find_or_initialize_by(:name => name).tap do |wf|
       wf.update!(
-        :name         => name,
-        :description  => description,
-        :manager_id   => manager_id,
-        :type         => self.class.module_parent::Workflow.name,
-        :payload      => payload,
-        :payload_type => "json"
+        :name          => name,
+        :description   => description,
+        :manager_id    => manager_id,
+        :type          => self.class.module_parent::Workflow.name,
+        :payload       => payload,
+        :payload_type  => "json",
+        :payload_valid => !!floe_workflow,
+        :payload_error => payload_error
       )
     end
   end
--- a/spec/models/manageiq/providers/workflows/automation_manager/configuration_script_source_spec.rb
+++ b/spec/models/manageiq/providers/workflows/automation_manager/configuration_script_source_spec.rb
@@ -206,18 +206,30 @@ RSpec.describe ManageIQ::Providers::Workflows::AutomationManager::ConfigurationS
       context "with workflows with invalid json" do
         let(:repo_dir_structure) { %w[invalid_json.asl] }
 
-        it "skips the invalid workflow file" do
+        it "sets the payload_valid and payload_error attributes" do
           record = build_record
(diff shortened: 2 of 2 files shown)


Fryguy:
I think it would be cool in the UI if you could see the payload_valid as a ✅ or ❌ in the table, then perhaps have the error on mouse over.  Or, at a minimum show the value on the summary page.

agrare:
I really want to write some specs for this, the way the fake repo is set up it makes it difficult to setup an invalid file for a single test.  Going to play with that a little bit

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -18,9 +18,24 @@ def sync
           next if filename.start_with?(".") || !filename.end_with?(".asl")
 
           payload = worktree.read_file(filename)
-          found   = current.delete(filename) || self.class.module_parent::Workflow.new(:configuration_script_source_id => id)
-
-          found.update!(:name => filename, :manager_id => manager_id, :payload => payload, :payload_type => "json")
+          payload_valid, payload_error =
+            begin
+              Floe::Workflow.new(payload)
```
Fryguy, line 23: If the payload is _very_ broken, like bad JSON broken, will it still raise a Floe::InvalidWorkflowError ?

agrare: Yes https://github.com/ManageIQ/floe/blob/master/lib/floe/workflow.rb#L57-L58

Fryguy:
I was just thinking - with the new column, we defaulted to valid (because I suggested it), but do we want a tri-boolean column? 😉...in all seriousness though, I could see valid, invalid and unknown, and the ansible ones could be unknown?  I started thinking about [my comment](https://github.com/ManageIQ/manageiq-providers-workflows/pull/59#issuecomment-1802502974) visualizing them in a table, and then I got to thinking how we want to collapse these views together and it might be weird to see ansible as "valid" even though we didn't verify them?  Thoughts?

agrare:
Hm would It be weird that `payload_valid?` on `nil` would be false when we just didn't check?
If we did have logic to say "valid", "invalid" or "unknown" would we show e.g. a grey question mark on the UI?  There's nothing the user could do to verify it though.

Fryguy:
Yeah I was thinking a gray question mark.  Alternatively we could always add some basic ansible validation.  (YAML parses, has a certain structure, etc)

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -33,4 +41,17 @@ def sync
...
+    git_repository.with_worktree do |worktree|
+      worktree.ref = scm_branch
+      worktree.blob_list.each do |filename|
+        next if filename.start_with?(".") || !filename.end_with?(".asl")
+
+        payload = worktree.read_file(filename)
+        yield filename, payload
+      end
+    end
+  end
```
agrare, line 56: Think this would be better in core's EmbeddedAutomationManager repository class but I wanted to test it out self-contained here.

kbrock:
Can we merge this even with valid column being true/false?

Fryguy (approved):
LGTM - just need to understand why tests are failing.

agrare:
Yeah we should probably disallow selection of invalid workflows in e.g. service template entrypoint lists also

PR #62, opened 2023-12-20
Update `:credentials` column if state sets a new credential

agrare:
If a state sets a new credential then encrypt the value and update the credentials column of the workflow instance.

NOTE: currently this will set a ManageIQ::Password encrypted value in the credentials jsonb column.  If there is an existing "mapped" authentication it will create a new in-line entry in the Credentials column instead of updating the Authentication record.

Documentation:
* https://github.com/ManageIQ/manageiq-documentation/pull/1773

Fix Auth:
* https://github.com/ManageIQ/manageiq/pull/22824


Fryguy:
Backported to `quinteros` in commit 733cad7562cdea1c302b3962b0b21aa5d10f2429.

```text
commit 733cad7562cdea1c302b3962b0b21aa5d10f2429
Date:   Thu Jan 4 16:51:17 2024 -0500

    Merge pull request #62 from agrare/allow_users_to_set_credentials_in_a_state
    
    Update `:credentials` column if state sets a new credential
    
    (cherry picked from commit 8c72b994824517eb0a66526be228f197e2e0950f)
```

PR #64, opened 2024-01-12
Handle the case where `:credentials` column is `nil`

agrare:
If the `:credentials` column is `nil` rather than `{}` as we had been testing with, the `resolved_credentials` method properly handles this but the `update_credentials!` method was not and caused the workflow instance run to throw an exception


Fryguy (approved):

Fryguy:
Backported to `quinteros` in commit 1046c124293c0c02c5e749ba7016b303b848f1a3.

```text
commit 1046c124293c0c02c5e749ba7016b303b848f1a3
Date:   Fri Jan 12 10:44:09 2024 -0500

    Merge pull request #64 from agrare/fix_update_credentials_nil
    
    Handle the case where `:credentials` column is `nil`
    
    (cherry picked from commit 41f5cde0d75e3d57833c890698719518008cfb37)
```

PR #65, opened 2024-01-17
Add the ability to set runner_options in Settings

agrare:
Depends on:
- [x] Release of floe with https://github.com/ManageIQ/floe/pull/155

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,0 +1,11 @@
+:ems:
+  :ems_workflows:
+    :runner_options:
+      :docker:
+        :network: host
+        :pull_policy: always
+      :kubernetes:
+        :pull_policy: Always
+      :podman:
+        :network: host
+        :pull_policy: always
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -38,6 +38,8 @@ module ManageIQ
           require "miq_environment"
           require "floe"
 
+          floe_runner_settings = Settings.ems.ems_workflows.runner_options
+
           if MiqEnvironment::Command.is_podified?
             host = ENV.fetch("KUBERNETES_SERVICE_HOST")
             port = ENV.fetch("KUBERNETES_SERVICE_PORT")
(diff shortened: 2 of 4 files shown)


lib/manageiq/providers/workflows/engine.rb
```diff
@@ -48,21 +48,26 @@ def self.floe_docker_runner
...
           elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
             options = {}
+
             if Rails.env.production?
               options["root"] = Rails.root.join("data/containers/storage").to_s
             else
               options["network"] = "host"
             end
 
+            options.merge!(Settings.ems.ems_workflows.runner_options.podman)
```
agrare, line 63: NOTE I wanted to merge after our defaults in case someone wanted to override one of these instead of doing `options = Settings.ems....` above then setting these defaults

config/settings/development.yml
```diff
@@ -0,0 +1,11 @@
+:ems:
+  :ems_workflows:
+    :runner_options:
+      :docker:
+        :network: host
```
agrare, line 5: Having a development.yml meant I could move the `options["network"] = "host" if Rails.env.development?` out of the engine :tada:

kbrock: huh. I never saw those implications. that is very cool

lib/manageiq/providers/workflows/engine.rb
```diff
@@ -48,21 +50,18 @@ def self.floe_docker_runner
...
 
             Floe::Workflow::Runner::Kubernetes.new(options)
           elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
             options = {}
-            if Rails.env.production?
-              options["root"] = Rails.root.join("data/containers/storage").to_s
-            else
-              options["network"] = "host"
-            end
+            options["root"] = Rails.root.join("data/containers/storage").to_s if Rails.env.production?
```
kbrock, line 58: the `development` part was so cool that my brain is working overtime to try and do the same thing with production.

PR #75, opened 2024-03-25
Add Settings.ems.ems_workflows.runner

kbrock:
a developer setting to run the various runners

### DIFF
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -37,13 +37,26 @@ module ManageIQ
           ]
         end
 
+        def self.floe_runner_name
+          if (runner_setting = Settings.ems.ems_workflows.runner.presence)
+            runner_setting
+          elsif MiqEnvironment::Command.is_podified?
+            "kubernetes"
+          elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
+            "podman"
+          else
+            "docker"
+          end
+        end
+
         def self.set_floe_runner
           require "miq_environment"
           require "floe"
 
           floe_runner_settings = Settings.ems.ems_workflows.runner_options
 
-          if MiqEnvironment::Command.is_podified?
+          case floe_runner_name
+          when "kubernetes"
             host = ENV.fetch("KUBERNETES_SERVICE_HOST")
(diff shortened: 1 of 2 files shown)


lib/manageiq/providers/workflows/engine.rb
```diff
@@ -34,13 +34,26 @@ def self.seedable_classes
           %w[ManageIQ::Providers::Workflows]
         end
 
+        def self.floe_runner_name
+          if (runner_setting = Settings.ems.ems_workflows.runner)
```
agrare, line 38: Wonder if we should we add `:runner:` to the `config/settings.yml` so we don't have hidden settings?

Fryguy: Agreed - we try to avoid hidden settings.

Fryguy: Or maybe we should add to config/environments/development.yml only?

I honestly don't understand the purpose of this setting.  It seems to imply that in dev mode we are not detecting the mode properly, which seems more of a bug to me.

agrare: This is actually something that I could use, on my workstation I have docker, podman, and kubernetes available to run with and I regularly test `floe` with all three.  It would be nice to override what the provider plugin uses.

I agree though `config/environments/development.yml` since you would never want to use this in production.

lib/manageiq/providers/workflows/engine.rb
```diff
@@ -53,13 +66,13 @@ def self.set_floe_runner
...
             Floe.set_runner("docker", "kubernetes", options)
-          elsif MiqEnvironment::Command.is_appliance? || MiqEnvironment::Command.supports_command?("podman")
+          when "podman"
             options = {}
             options["root"] = Rails.root.join("data/containers/storage").to_s if Rails.env.production?
             options.merge!(floe_runner_settings.podman)
 
             Floe.set_runner("docker", "podman", options)
-          else
+          else # "docker"
```
agrare, line 75: @kbrock maybe overkill but what do you think about `when "docker"` and the `else` condition handling if you entered an invalid runner name in settings?  The way this is, if you typo'd `pdoman` it'd default to docker.

kbrock:
update:
- handle case where developer configures a bad docker runner

agrare (approved):

PR #76, opened 2024-03-29
Set workflow description from asl comment

agrare:
We do something similar if you manually create a Workflow from a json payload since we don't have a file path, https://github.com/ManageIQ/manageiq-providers-workflows/blob/master/app/models/manageiq/providers/workflows/automation_manager/workflow.rb#L4 and I thought it'd add a nice "human readable" description

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -76,8 +76,24 @@ class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSour
   end
 
   def create_workflow_from_payload(name, payload)
+    floe_workflow =
+      begin
+        Floe::Workflow.new(payload)
+      rescue Floe::InvalidWorkflowError
+        nil
+      end
+
+    description = floe_workflow&.comment
+
     configuration_script_payloads.find_or_initialize_by(:name => name).tap do |wf|
-      wf.update!(:name => name, :manager_id => manager_id, :type => self.class.module_parent::Workflow.name, :payload => payload, :payload_type => "json")
+      wf.update!(
+        :name         => name,
+        :description  => description,
+        :manager_id   => manager_id,
+        :type         => self.class.module_parent::Workflow.name,
+        :payload      => payload,
+        :payload_type => "json"
+      )
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -76,8 +76,26 @@ def sync_from_content(to_delete)
   end
 
   def create_workflow_from_payload(name, payload)
+    floe_workflow, payload_error =
+      begin
+        Floe::Workflow.new(payload)
```
agrare, line 81: I had this depend on https://github.com/ManageIQ/manageiq-providers-workflows/pull/59 because we were already loading the Floe::Workflow there

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -76,8 +76,26 @@ def sync_from_content(to_delete)
...
+      rescue Floe::InvalidWorkflowError => err
+        [nil, err.message]
+      end
+
+    comment = floe_workflow&.payload&.[]("Comment")
```
agrare, line 86: TODO I think having `Floe::Workflow#comment` accessor would be cleaner

kbrock: => https://github.com/ManageIQ/floe/pull/170

kbrock:
Can we merge this without the sync PR?
Seems like the schema PR is on hold for a bit

app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -76,8 +76,26 @@ def sync_from_content(to_delete)
...
+
+    description = floe_workflow&.comment
+
     configuration_script_payloads.find_or_initialize_by(:name => name).tap do |wf|
-      wf.update!(:name => name, :manager_id => manager_id, :type => self.class.module_parent::Workflow.name, :payload => payload, :payload_type => "json")
+      wf.update!(
+        :name          => name,
+        :description   => description,
+        :manager_id    => manager_id,
+        :type          => self.class.module_parent::Workflow.name,
```
kbrock, line 93: I'm surprised we are changing the `type`. (for anything)

agrare:
Okay rebased to not depend on https://github.com/ManageIQ/manageiq-providers-workflows/pull/59

PR #78, opened 2024-04-25
[WIP] [PoC] Add Builtin Runner

Fryguy:

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_runner.rb
+++ b/lib/manageiq/providers/workflows/builtin_runner.rb
@@ -1,0 +1,41 @@
+module ManageIQ
+  module Providers
+    module Workflows
+      class BuiltinRunnner < Floe::Runner
+        SCHEME = "builtin"
+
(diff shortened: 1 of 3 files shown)


lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        def wait(timeout: nil, events: %i[create update delete])
+          # TODO: wait_for_task?
+        end
+
+        def status!(runner_context)
+        end
+
+        def running?(runner_context)
+        end
```
agrare, line 29: If some of the builtin methods we are going to call will be async we can use this method here to control when we call  back into the workflow to step through the next state.

We could define a `_ready?` method for each builtin that will indicate when it is complete (can just be `true` for fire-and-forget methods).

Fryguy:
Closing as this is being explored further in other PRs.

PR #81, opened 2024-05-02
Add builtin method: provision_execute

kbrock:
dependency:
- [x] https://github.com/ManageIQ/manageiq-providers-workflows/pull/85
- [x] https://github.com/ManageIQ/manageiq/pull/23021

status: sharing my current code

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -16,7 +16,7 @@ module ManageIQ
 
         private_class_method def self.miq_task_status!(runner_context)
           miq_task = ::MiqTask.find_by(:id => runner_context["miq_task_id"])
-          return error!(runner_context, :cause => "Unable to find MiqTask id: [#{runner_context["miq_task_id"]}]") if miq_task.nil?
+          return BuiltinRunner.error!(runner_context, :cause => "Unable to find MiqTask id: [#{runner_context["miq_task_id"]}]") if miq_task.nil?
 
           runner_context["running"] = miq_task.state != ::MiqTask::STATE_FINISHED
 
@@ -25,17 +25,11 @@ module ManageIQ
             if runner_context["success"]
               runner_context["output"] = miq_task.message
             else
-              error!(runner_context, :cause => miq_task.message)
+              BuiltinRunner.error!(runner_context, :cause => miq_task.message)
             end
           end
 
           runner_context
-        end
-
-        private_class_method def self.error!(runner_context = {}, cause:, error: "States.TaskFailed")
-          runner_context.merge!(
-            "running" => false, "success" => false, "output" => {"Error" => error, "Cause" => cause}
-          )
         end
       end
     end
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -14,6 +14,24 @@ module ManageIQ
           miq_task_status!(runner_context)
         end
 
+        def self.provision_execute(_params = {}, _secrets = {}, context = {})
+          # ensure we are in a provisioning request
+          unless context.execution["_object_type"] == "ServiceTemplateProvisionTask"
+            return BuiltinRunner.error!(runner_context, :cause => "Calling provision_execute on non-provisioning request: [#{runner_context["_object_type"]}]")
+          end
+
+          request_task_id = context.execution["_object_id"]&.to_i
+          ::MiqRequestTask.find(request_task_id)&.execute
+
+          {"miq_request_task_id" => request_task_id}
+        end
+
+        private_class_method def self.provision_execute_status!(runner_context)
+          miq_request_task_status!(runner_context)
+        end
+
+        # general methods
+
         private_class_method def self.miq_task_status!(runner_context)
           miq_task = ::MiqTask.find_by(:id => runner_context["miq_task_id"])
           return BuiltinRunner.error!(runner_context, :cause => "Unable to find MiqTask id: [#{runner_context["miq_task_id"]}]") if miq_task.nil?
@@ -30,6 +48,24 @@ module ManageIQ
           end
 
           runner_context
+        end
+
+        private_class_method def self.miq_request_task_status!(runner_context)
+          miq_request_task = ::MiqRequestTask.find_by(:id => runner_context["miq_request_task_id"])
+
+          case miq_request_task&.statemachine_task_status
+          when nil
+            reason = "Unable to find MiqRequestTask id: [#{runner_context["miq_request_task_id"]}]"
+            BuiltinRunner.error!(runner_context, :cause => reason)
+          when "error"
+            reason = request_task.message&.sub(/^Error: /, "")
+            BuiltinRunner.error!(runner_context, :cause => reason)
+          when "retry"
+            runner_context["running"] = true
+            runner_context
+          when "ok"
+            BuiltinRunner.success!(runner_context, :output => {"Result" => "provisioned"})
+          end
         end
       end
     end
--- a/lib/manageiq/providers/workflows/builtin_runner.rb
+++ b/lib/manageiq/providers/workflows/builtin_runner.rb
@@ -55,6 +55,12 @@ module ManageIQ
             "running" => false, "success" => false, "output" => {"Error" => error, "Cause" => cause}
           )
         end
+
+        def self.success!(runner_context = {}, output:)
+          runner_context.merge!(
+            "running" => false, "success" => true, "output" => output
+          )
+        end
       end
     end
   end
--- a/lib/manageiq/providers/workflows/builtin_runner.rb
+++ b/lib/manageiq/providers/workflows/builtin_runner.rb
@@ -10,13 +10,13 @@ module ManageIQ
           scheme_prefix = "#{SCHEME}://"
           raise ArgumentError, "Invalid resource" unless resource&.start_with?(scheme_prefix)
 
-          method = resource.sub(scheme_prefix, "")
+          method_name = resource.sub(scheme_prefix, "")
 
-          runner_context = {"method" => method}
+          runner_context = {"method" => method_name}
 
           # TODO: prevent calling anything except the specifics methods, e.g. you shouldn't be able to call .to_s.
           #       Maybe make BuiltinMethods a BasicObject?
-          method_result = BuiltinMethods.public_send(method, params, secrets, context)
+          method_result = BuiltinMethods.public_send(method_name, params, secrets, context)
           method_result.merge(runner_context)
         rescue => err
           cleanup(runner_context)
@@ -24,10 +24,10 @@ module ManageIQ
         end
 
         def cleanup(runner_context)
-          method = runner_context["method"]
-          raise ArgumentError if method.nil?
+          method_name = runner_context["method"]
+          raise ArgumentError if method_name.nil?
 
-          cleanup_method = "#{method}_cleanup"
+          cleanup_method = "#{method_name}_cleanup"
           return unless BuiltinMethods.respond_to?(cleanup_method, true)
 
           BuiltinMethods.send(cleanup_method, runner_context)
@@ -38,10 +38,10 @@ module ManageIQ
         end
 
         def status!(runner_context)
-          method = runner_context["method"]
-          raise ArgumentError if method.nil?
+          method_name = runner_context["method"]
+          raise ArgumentError if method_name.nil?
 
-          BuiltinMethods.send("#{method}_status!", runner_context)
+          BuiltinMethods.send("#{method_name}_status!", runner_context)
         end
 
         def running?(runner_context)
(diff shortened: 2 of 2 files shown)


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -21,6 +21,43 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+        end
+
+        private_class_method def self.provision_execute_status!(runner_context)
+          request_task = ::MiqRequestTask.find_by(:id => runner_context["miq_request_task_id"])
+          return if request_task.nil?
+
+          case request_task.statemachine_task_status
+          when "error"
+            reason = task.message
+            reason = reason[7..] if reason[0..6] == "Error: "
```
Fryguy, line 41:
```suggestion
            reason = reason[7..] if reason.start_with?("Error: ")
```

or, perhaps just use sub which I think is more readable

```suggestion
            reason = reason.sub(/^Error: ", "")
```

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -21,6 +21,43 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+            runner_context["output"] = {"Error" => "States.TaskFailed", "Cause" => reason}
+          when "retry"
+            runner_context["running"] = false
+            runner_context["success"] = false
+            # TODO: put in proper error to enable retrying. Maybe timeout?
+            runner_context["output"] = {"Error" => "States.Timeout", "Cause" => "retry"}
+          when "ok"
+            runner_context["running"] = false
+            runner_context["success"] = false
+            runner_context["output"] = {"Result" => "provisioned"}
```
Fryguy, line 53: Should there be an else case?

Fryguy (dismissed):
Overall LGTM.  @agrare Can you please also review?

Fryguy:
@kbrock This PR was titled "Builtin runner accepting context", but then it also includes provision_execute and some variable renames and builtin flow changes - was this supposed to be separate PRs or should the title be renamed to something like "Add provision_execute as a builtin method"?

Fryguy:
I just realized this requires https://github.com/ManageIQ/manageiq/pull/23021, so I updated the OP

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -21,6 +21,43 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+            runner_context["success"] = false
+            runner_context["output"] = {"Error" => "States.TaskFailed", "Cause" => reason}
+          when "retry"
+            runner_context["running"] = false
+            runner_context["success"] = false
+            # TODO: put in proper error to enable retrying. Maybe timeout?
+            runner_context["output"] = {"Error" => "States.Timeout", "Cause" => "retry"}
+          when "ok"
+            runner_context["running"] = false
+            runner_context["success"] = false
```
agrare, line 52: Doesn't `"ok"` mean that success should be true here?

lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -49,6 +54,9 @@ def success?(runner_context)
         def output(runner_context)
           runner_context["output"]
         end
+
+        # def wait(_timeout: nil, _events: %i[create update delete])
+        # end
```
agrare, line 59:
```suggestion
```

lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -34,8 +38,9 @@ def wait(timeout: nil, events: %i[create update delete])
         def status!(runner_context)
           method = runner_context["method"]
           raise ArgumentError if method.nil?
+          return if !runner_context["running"]
 
-          BuiltinMethods.send("#{method}_status!", runner_context)
+          BuiltinMethods.send(status_method, runner_context)
```
Fryguy, line 43: Where is this `status_method` defined?

lib/manageiq/providers/workflows/builtin_runner.rb
kbrock: ?

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -21,6 +21,41 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+          return if request_task.nil?
+
+          case request_task.statemachine_task_status
+          when "error"
+            reason = task.message.sub(/^Error: /, "")
+            runner_context["running"] = false
+            runner_context["success"] = false
+            runner_context["output"] = {"Error" => "States.TaskFailed", "Cause" => reason}
+          when "retry"
+            runner_context["running"] = false
```
Fryguy, line 44: I'm not sure what retry implies, but I think it should still be "running"?

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -21,6 +21,41 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+          return unless context.execution["_object_type"] == "ServiceTemplateProvisionTask"
+
+          ::MiqRequestTask.find(context.execution["_object_id"])&.execute
+
+          {"miq_request_task_id" => request_task_id}
+        end
+
+        private_class_method def self.provision_execute_status!(runner_context)
+          request_task = ::MiqRequestTask.find_by(:id => runner_context["miq_request_task_id"])
+          return if request_task.nil?
```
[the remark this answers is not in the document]
Fryguy: Regardless, the caller does `method_result.merge(runner_context)`, so this will just cause the caller to blow up with a `undefined method `merge' for nil:NilClass`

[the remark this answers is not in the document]
agrare: Right I'm thinking we should return `{"running" => false, "success" => "false", "output" => {"Error" => "States.TaskFailed", "Cause" => "Unable to find miq_request_task"}}` or something like that

lib/manageiq/providers/workflows/builtin_methods.rb
agrare: Oh yeah good point, does "retry" here mean we want the ASL workflow to retry or it is the backend doing the retry?

kbrock: I was thinking that it was the same thing as a failure, the difference being that we will revisit this.

see how this is written in the floe:
https://github.com/ManageIQ/manageiq-content/pull/746

I agree, this suggests that it is a retry, but this version of `retry` may be more like a `wait`.

lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -27,13 +27,10 @@ def cleanup(runner_context)
...
         end
 
-        def wait(timeout: nil, events: %i[create update delete])
-          # TODO: wait_for_task?
-        end
-
         def status!(runner_context)
           method = runner_context["method"]
           raise ArgumentError if method.nil?
+          return if !runner_context["running"]
```
Fryguy, line 33: I'm not sure I understand this bit - Is this to prevent status being called twice?

lib/manageiq/providers/workflows/builtin_methods.rb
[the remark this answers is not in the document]
Fryguy: Oh no wait, I thought this was the execute method as opposed to the status method.  Ignore my comment.

lib/manageiq/providers/workflows/builtin_runner.rb
kbrock: yes. status! is called 2-3 times once a task is complete

also, if a task fails in the original run, then we don't want it calling status

agrare: > also, if a task fails in the original run, then we don't want it calling status

In the container_runner classes we return if an error has already been set in the runner context, would that work for this case?

kbrock: The check on Error was the inspiration.
The thinking was that if we set the `output` / `running=false`, then we were done with this task and no reason to run status to check further

lib/manageiq/providers/workflows/builtin_methods.rb
kbrock: confused by this.
agree with `running=true`, but it is not complete. it is basically a `wait` state so then `success` is not fully determined.

agrare: > agree with running=true, but it is not complete.

Do you mean you don't think we should set success=true?

This could apply to all of these, but in general the container_runners don't set success=true/false until after it has completed so if that's what you mean then yes I agree but then still not sure why we'd say running=false,success=false because that is an error case.

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -21,6 +21,41 @@ def self.email(params = {}, _secrets = {}, _context = {})
 
         private_class_method def self.email_cleanup(*)
         end
+
+        def self.provision_execute(_params = {}, _secrets = {}, context = {})
+          return unless context.execution["_object_type"] == "ServiceTemplateProvisionTask"
+
+          ::MiqRequestTask.find(context.execution["_object_id"])&.execute
+
+          {"miq_request_task_id" => request_task_id}
```
agrare, line 30:
```
#<NameError: undefined local variable or method `request_task_id' for ManageIQ::Providers::Workflows::BuiltinMethods:Class

          {"miq_request_task_id" => request_task_id}
```

kbrock:
update
- fixed miq_request_task_id
- fixed delay
- `s/method/method_name` to avoid using common method name

lib/manageiq/providers/workflows/builtin_methods.rb
Fryguy: I was thinking running=true and success is not yet determined (not sure if that should be false or just not set at all?)

lib/manageiq/providers/workflows/builtin_methods.rb
Fryguy: Think this is available in line 28 now? (old comment?)

agrare: This comment was from before the most recent push which fixed it

lib/manageiq/providers/workflows/builtin_methods.rb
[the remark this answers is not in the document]
kbrock: this nil is copied from email status.
I'll fix, but maybe we want to consider fixing that other one as well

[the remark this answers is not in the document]
agrare: >  I'll fix, but maybe we want to consider fixing that other one as well

Yes I think I commented over in the PR that added that but didn't get to fix it, wanted to get it right here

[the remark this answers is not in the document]
agrare: @kbrock https://github.com/ManageIQ/manageiq-providers-workflows/pull/87/files#diff-f49bf1f6a21ddda0e802d7717ef72ecaef1b11b86ee27fb5db029c7f770e3b0cR19

kbrock:
update:
- split into multiple commits
- extract status check for miq_request_task
- move error and status from methods to runner

update:
- fixed `case when nil` case

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -14,9 +14,27 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
         end
 
+        def self.provision_execute(_params = {}, _secrets = {}, context = {})
+          # ensure we are in a provisioning request
+          unless context.execution["_object_type"] == "ServiceTemplateProvisionTask"
+            return BuiltinRunner.error!(runner_context, :cause => "Calling provision_execute on non-provisioning request: [#{runner_context["_object_type"]}]")
+          end
+
+          request_task_id = context.execution["_object_id"]&.to_i
+          ::MiqRequestTask.find(request_task_id)&.execute
```
agrare, line 24: @kbrock it looks like we go through some effort to ensure `context.execution["_object_type"]` is `"ServiceTemplateProvisionTask"` but then also go through a lot of effort to handle a missing `_object_id`.

Should we also be raising an error if `context.execution["_object_type"]` is missing and if `::MiqRequestTask.find(request_task_id)` can't be found?

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -25,17 +43,29 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+            reason = "Unable to find MiqRequestTask id: [#{runner_context["miq_request_task_id"]}]"
+            BuiltinRunner.error!(runner_context, :cause => reason)
+          when "error"
+            reason = request_task.message&.sub(/^Error: /, "")
+            BuiltinRunner.error!(runner_context, :cause => reason)
+          when "retry"
+            runner_context["running"] = true
+            runner_context
+          when "ok"
+            BuiltinRunner.success!(runner_context, :output => {"Result" => "provisioned"})
```
agrare, line 67: We should probably return something that subsequent states could make use of, maybe the resulting `MiqRequestTask#destination` which is a `Service`

Example use case, run some automation against the vm(s) that get created.

agrare (approved):
Just a couple of nits that we can do in a follow-up

PR #83, opened 2024-05-02
Workflow#run accepts context

kbrock:
dependency:
- tag: v0.11.0 (already updated)

pass context into run.
This allows us to pass `_miq_request_task_id` for provisioning (and other) requests

we could probably drop the `role` parameter

Feedback:
- Do we want to set `object.class, object.id`?


agrare:
> we could probably drop the role parameter

I added this in case we want to run workflows for internal purposes similar to how we use embedded_ansible for some provider operations

agrare (approved):

PR #84, opened 2024-05-03
Move checking builtin method status into Methods

agrare:
Move the logic for checking if a method is running/successful into the BuiltinMethods so that each method can control how it checks.

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -9,6 +9,18 @@ module ManageIQ
 
           {"miq_task_id" => miq_task.id}
         end
+
+        private_class_method def self.email_status!(runner_context)
+          miq_task = ::MiqTask.find(runner_context["miq_task_id"])
+          return if miq_task.nil?
+
+          runner_context["running"] = miq_task.state != ::MiqTask::STATE_FINISHED
+          runner_context["success"] = miq_task.status == ::MiqTask::STATUS_OK unless runner_context["running"]
+          runner_context
+        end
+
+        private_class_method def self.email_cleanup(*)
+        end
       end
     end
   end
--- a/lib/manageiq/providers/workflows/builtin_runner.rb
+++ b/lib/manageiq/providers/workflows/builtin_runner.rb
@@ -12,12 +12,19 @@ module ManageIQ
 
           method = resource.sub(scheme_prefix, "")
 
+          runner_context = {"method" => method}
+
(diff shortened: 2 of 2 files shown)


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -9,6 +9,18 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
           {"miq_task_id" => miq_task.id}
         end
+
+        private_class_method def self.email_status!(runner_context)
+          miq_task = ::MiqTask.find(runner_context["miq_task_id"])
+          return if miq_task.nil?
+
+          runner_context["running"] = miq_task.state != ::MiqTask::STATE_FINISHED
+          runner_context["success"] = miq_task.status == ::MiqTask::STATUS_OK unless runner_context["running"]
+          runner_context
```
agrare, line 19: This can probably be a `miq_task_status!`

lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -12,31 +12,30 @@ def run_async!(resource, params = {}, secrets = {}, context = {})
 
           method = resource.sub(scheme_prefix, "")
 
+          runner_context = {"method" => method}
```
agrare, line 15: I had to add this so that we know what method was called in the subsequent calls down below.

I'd like to tweak how Floe Task is calling the runner so maybe these are instance methods with the resource as an instance variable so that this is a little cleaner

lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -12,31 +12,30 @@ def run_async!(resource, params = {}, secrets = {}, context = {})
...
           #       Maybe make BuiltinMethods a BasicObject?
-          BuiltinMethods.public_send(method, params, secrets, context)
+          method_result = BuiltinMethods.public_send(method, params, secrets, context)
+          method_result.merge(runner_context)
         end
 
-        def cleanup(_runner_context)
+        def cleanup(runner_context)
+          method = runner_context["method"]
+          raise ArgumentError if method.nil?
```
agrare, line 25: This just catches if someone calls e.g. cleanup before run_async!

Fryguy:
Overall LGTM.  I agree on your comments above, but I don't want to hold it up because those just make it cleaner, where this actually starts defining the API.

Fryguy:
Changed this to enhancement because it also adds the method def shape as the API.

PR #86, opened 2024-05-07
[WIP] Add an embedded_ansible builtin method

agrare:


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -11,15 +11,53 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+
+          vars = params
+                 .slice("Hosts", "ExtraVars", "BecomeEnabled", "Timeout", "Verbosity", "Credential", "CloudCredential", "NetworkCredential", "VaultCredential")
+                 .transform_keys { |k| k.underscore.to_sym }
+
+          repository = ::ConfigurationScriptSource.find_by(:scm_url => repository_url, :scm_branch => repository_branch)
+          return error!(runner_context, :cause => "Unable to find Repository: URL: [#{repository_url}] Branch: [#{repository_branch}]") if repository.nil?
+
+          playbook = ::ConfigurationScriptPayload.find_by(:configuration_script_source => repository, :name => playbook)
+          return error!(runner_context, :cause => "Unable to find Playbook: Name: [#{playbook}] Repository: #{repository.name}") if playbook.nil?
```
agrare, line 28: NOTE I thought that looking up a playbook by repository_url/repository_branch+playbook_path was the most unique way to identify a playbook without getting into IDs which obviously wouldn't work across multiple miq instances.

Fryguy: I wonder if we can support both, but yeah I agree this is a good way to start it.

agrare: Yeah I was thinking of adding a PlaybookId and you have to specify either PlaybookId or RepoUrl+RepoBranch+PlaybookPath

agrare: The non ID way would be better for service import/export as well

agrare:
> Are these id's? If so, should these be something like

I was keeping the argument names from https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/playbook.rb#L36-L43 wherever possible, I don't have an issue changing it if it is better for the user but for now I just take exactly what `Playbook#run` accepts as args

agrare:
@Fryguy okay took care of the TODO items please take a look

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -14,6 +14,39 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+        def self.embedded_ansible(params = {}, _secrets = {}, _context = {})
+          repository_url, repository_branch, playbook_name, playbook_id = params.values_at("RepositoryUrl", "RepositoryBranch", "PlaybookName", "PlaybookId")
+
+          vars = params
+                 .slice("Hosts", "ExtraVars", "BecomeEnabled", "Timeout", "Verbosity", "CredentialId", "CloudCredentialId", "NetworkCredentialId", "VaultCredentialId")
+                 .transform_keys { |k| k.underscore.to_sym }
+
+          vars[:execution_ttl] = vars.delete(:timeout) if vars.key?(:timeout)
+          %i[credential_id cloud_credential_id network_credential_id vault_credential_id].each do |key|
+            new_key = key.to_s.gsub(/_id$/, '').to_sym
```
Fryguy, line 26: Not necessary to fix, but this can also be simplified to
```suggestion
            new_key = key.to_s.chomp("_id").to_sym
```

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -14,6 +14,39 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+
+          if playbook_id
+            playbook = ::ConfigurationScriptPayload.find_by(:id => playbook_id)
+            return BuiltinRunnner.error!({}, :cause => "Unable to find Playbook: Id: [#{playbook_id}] Repository: [#{repository.name}]") if playbook.nil?
+          else
+            repository = ::ConfigurationScriptSource.find_by(:scm_url => repository_url, :scm_branch => repository_branch)
+            return BuiltinRunnner.error!({}, :cause => "Unable to find Repository: URL: [#{repository_url}] Branch: [#{repository_branch}]") if repository.nil?
+
+            playbook = ::ConfigurationScriptPayload.find_by(:configuration_script_source => repository, :name => playbook_name)
+            return BuiltinRunnner.error!({}, :cause => "Unable to find Playbook: Name: [#{playbook_name}] Repository: [#{repository.name}]") if playbook.nil?
```
Fryguy, line 38: These are going to find any ConfigurationScriptSource.  Should we limit this to embedded ansible specific subclasses?

Fryguy (approved):
LGTM - going to merge as is, and we can do anything else in follow ups.

Fryguy:
Backported to `radjabov` in commit 481905594984e36dbb91746b78fa7303fc82b270.

```text
commit 481905594984e36dbb91746b78fa7303fc82b270
Date:   Thu Jun 6 12:12:58 2024 -0400

    Merge pull request #86 from agrare/add_embedded_ansible_builtin_method
    
    Add an embedded_ansible builtin method
    
    (cherry picked from commit 6d2221abaa46418a1de177de92ddeeee56f5474b)
```

PR #87, opened 2024-05-07
Some minor refactorings to BuiltinMethods / BuiltinRunner

agrare:
1. Check private methods for `BuiltinRunner.respond_to?`
2. Extract `miq_task_status!` so we can reuse it for other methods
3. Add an `error_payload` method to make returning errors simpler

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -15,21 +15,27 @@ module ManageIQ
         end
 
         private_class_method def self.miq_task_status!(runner_context)
-          miq_task = ::MiqTask.find(runner_context["miq_task_id"])
-          return error_payload(:cause => "Unable to find MiqTask id: [#{runner_context["miq_task_id"]}]") if miq_task.nil?
+          miq_task = ::MiqTask.find_by(:id => runner_context["miq_task_id"])
+          return error!(runner_context, :cause => "Unable to find MiqTask id: [#{runner_context["miq_task_id"]}]") if miq_task.nil?
 
           runner_context["running"] = miq_task.state != ::MiqTask::STATE_FINISHED
 
           unless runner_context["running"]
             runner_context["success"] = miq_task.status == ::MiqTask::STATUS_OK
-            runner_context["output"]  = runner_context["success"] ? miq_task.message : {"Error" => "States.TaskFailed", "Cause" => miq_task.message}
+            if runner_context["success"]
+              runner_context["output"] = miq_task.message
+            else
+              error!(runner_context, :cause => miq_task.message)
+            end
           end
 
           runner_context
         end
 
-        private_class_method def self.error_payload(cause:, error: "States.TaskFailed")
-          {"running" => false, "success" => false, "output" => {"Error" => error, "Cause" => cause}}
+        private_class_method def self.error!(runner_context = {}, cause:, error: "States.TaskFailed")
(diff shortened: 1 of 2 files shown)


kbrock:
good. I just added/deleted an `error_payload` method.
probably created it 5 times but was debating. thanks

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -11,15 +11,25 @@ def self.email(params = {}, _secrets = {}, _context = {})
         end
 
         private_class_method def self.email_status!(runner_context)
+          miq_task_status!(runner_context)
+        end
+
+        private_class_method def self.miq_task_status!(runner_context)
           miq_task = ::MiqTask.find(runner_context["miq_task_id"])
```
kbrock, line 18: I think `find` throws an exception if the record is not found

I have this change locally:

```suggestion
          miq_task = ::MiqTask.find_by(:id => runner_context["miq_task_id"])
```

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -11,15 +11,25 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
+            runner_context["success"] = miq_task.status == ::MiqTask::STATUS_OK
+            runner_context["output"]  = runner_context["success"] ? miq_task.message : {"Error" => "States.TaskFailed", "Cause" => miq_task.message}
+          end
+
           runner_context
         end
 
-        private_class_method def self.email_cleanup(*)
+        private_class_method def self.error_payload(cause:, error: "States.TaskFailed")
+          {"running" => false, "success" => false, "output" => {"Error" => error, "Cause" => cause}}
```
kbrock, line 32: the method `status!` is modifying a context in place.
Lets modify it rather than generating a new one.
```suggestion
        private_class_method def self.error_payload(runner_context = {}, cause:, error: "States.TaskFailed")
          runner_context.merge("running" => false, "success" => false, "output" => {"Error" => error, "Cause" => cause})
```

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -11,15 +11,25 @@ def self.email(params = {}, _secrets = {}, _context = {})
...
-          return if miq_task.nil?
+          return error_payload(:cause => "Unable to find MiqTask id: [#{runner_context["miq_task_id"]}]") if miq_task.nil?
 
           runner_context["running"] = miq_task.state != ::MiqTask::STATE_FINISHED
-          runner_context["success"] = miq_task.status == ::MiqTask::STATUS_OK unless runner_context["running"]
+
+          unless runner_context["running"]
+            runner_context["success"] = miq_task.status == ::MiqTask::STATUS_OK
+            runner_context["output"]  = runner_context["success"] ? miq_task.message : {"Error" => "States.TaskFailed", "Cause" => miq_task.message}
+          end
```
kbrock, line 26: do we want to merge this with `error_payload`?

kbrock (commented):
mainly concerned about modifying the context vs returning a new one.

lib/manageiq/providers/workflows/builtin_methods.rb
agrare: Yeah possibly, I considered doing `runner_context = error_payload...` but that would overwrite rather than merge with the argument passed in

PR #89, opened 2024-05-08
Add specs for BuiltinMethods

agrare:

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -18,11 +18,11 @@ module ManageIQ
           object_type, object_id = context.execution.values_at("_object_type", "_object_id")
 
           # ensure we are in a provisioning request
-          return BuiltinRunner.error!(runner_context, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless object_type == "ServiceTemplateProvisionTask"
-          return BuiltinRunner.error!(runner_context, :cause => "Missing MiqRequestTask id") if object_id.nil?
+          return ::ManageIQ::Providers::Workflows::BuiltinRunnner.error!({}, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless object_type == "ServiceTemplateProvisionTask"
+          return ::ManageIQ::Providers::Workflows::BuiltinRunnner.error!({}, :cause => "Missing MiqRequestTask id") if object_id.nil?
 
           miq_request_task = ::MiqRequestTask.find_by(:id => object_id.to_i)
-          return BuiltinRunner.error!(runner_context, :cause => "Unable to find MiqReqeustTask id: [#{object_id}]") if miq_request_task.nil?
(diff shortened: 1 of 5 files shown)


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -18,11 +18,11 @@ def self.provision_execute(_params = {}, _secrets = {}, context = {})
           object_type, object_id = context.execution.values_at("_object_type", "_object_id")
 
           # ensure we are in a provisioning request
-          return BuiltinRunner.error!(runner_context, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless object_type == "ServiceTemplateProvisionTask"
-          return BuiltinRunner.error!(runner_context, :cause => "Missing MiqRequestTask id") if object_id.nil?
+          return ::ManageIQ::Providers::Workflows::BuiltinRunnner.error!({}, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless object_type == "ServiceTemplateProvisionTask"
```
agrare, line 21: This is pretty annoying, not even `module_parent::BuiltinRunner` works

Fryguy: that's so weird - I'm sure it's a BasicObject + zeitwerk problem - maybe we need to manually require it ahead of time?

Fryguy:
Looks good - I'd be ok with merging this as is and we can iterate on the specs.

agrare:
@Fryguy let me get a spec covering the BuiltinRunner methods to make sure we're good there then yeah I agree we can work on adding more specs as we go

lib/manageiq/providers/workflows/builtin_runner.rb
```diff
@@ -12,12 +16,17 @@ def run_async!(resource, params = {}, secrets = {}, context = {})
...
-          method_result.merge(runner_context)
-        rescue => err
-          cleanup(runner_context)
-          error!(runner_context, :cause => err.to_s)
+          begin
+            runner_context = {"method" => method_name}
+            method_result = BuiltinMethods.public_send(method_name, params, secrets, context)
+            method_result.merge(runner_context)
+          rescue NoMethodError
+            error!(runner_context, :cause => "undefined method [#{method_name}]")
```
agrare, line 24: NOTE I added this specifically because I think that getting the method name wrong is going to be a very likely error and the payload that was returned was very much "internal ruby stuff" that would confuse a workflow author

```
{"method"=>"invalid_method",
 "running"=>false,
 "success"=>false,
 "output"=>
  {"Error"=>"States.TaskFailed",
   "Cause"=>
    "undefined method `invalid_method' for ManageIQ::Providers::Workflows::BuiltinMethods:Class\n" +
    "\n" +
    "            method_result = BuiltinMethods.public_send(method_name, params, secrets, context)\n" +
    "                                          ^^^^^^^^^^^^"}}
```

Fryguy (approved):

PR #92, opened 2024-05-13
[WIP] Only allow provision execute MiqProvision objects

agrare:


kbrock:
I was expecting the embedded_ansible to be over in #86

agrare:
> I was expecting the embedded_ansible to be over in https://github.com/ManageIQ/manageiq-providers-workflows/pull/86

Yeah this was built on that or to make it easier for the demo.

PR #94, opened 2024-05-21
automatically update options for provisioning requests

kbrock:
Use input to set options in the provisioning request.
This allows a workflow to manipulate the request.

- We always change the miq_request_task, even when the options have not been updated. Optionally we can add a sync parameter
- For the 2 non symbol keys, we do end up with duplicates (`nil` -> `""`, and the fqname)

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -23,6 +23,10 @@ module ManageIQ
           return BuiltinRunnner.error!({}, :cause => "Unable to find MiqReqeustTask id: [#{object_id}]")                        if miq_request_task.nil?
           return BuiltinRunnner.error!({}, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless miq_request_task.class < ::MiqProvision
 
+          new_options = context.input.symbolize_keys.slice(*miq_request_task.options.keys)
+          miq_request_task.options_will_change!
+          miq_request_task.options.merge!(new_options)
+          miq_request_task.save!
           miq_request_task.execute_queue
 
           {"miq_request_task_id" => miq_request_task.id}
--- a/spec/lib/manageiq/providers/workflows/builtin_methods_spec.rb
+++ b/spec/lib/manageiq/providers/workflows/builtin_methods_spec.rb
@@ -17,14 +17,49 @@ RSpec.describe ManageIQ::Providers::Workflows::BuiltinMethods do
     let(:params)  { {} }
     let(:secrets) { {} }
 
-    it "returns an error if _object_type isn't passed" do
-      runner_context = described_class.provision_execute(params, secrets, Floe::Workflow::Context.new({"Execution" => {"_object_id" => nil}}))
-      expect(runner_context).to include("running" => false, "success" => false, "output" => {"Error" => "States.TaskFailed", "Cause" => "Missing MiqRequestTask type"})
+    it "requires _object_type" do
+      runner_context = described_class.provision_execute(params, secrets, create_floe_context(:execution => {"_object_id" => nil}))
+      expect(runner_context).to include("running" => false, "success" => false, "output" => failed_task_status("Missing MiqRequestTask type"))
     end
 
-    it "returns an error if _object_id isn't passed" do
-      runner_context = described_class.provision_execute(params, secrets, Floe::Workflow::Context.new({"Execution" => {"_object_type" => "MiqProvision"}}))
-      expect(runner_context).to include("running" => false, "success" => false, "output" => {"Error" => "States.TaskFailed", "Cause" => "Missing MiqRequestTask id"})
+    it "requires _object_id" do
+      runner_context = described_class.provision_execute(params, secrets, create_floe_context(:execution => {"_object_type" => "MiqProvision"}))
+      expect(runner_context).to include("running" => false, "success" => false, "output" => failed_task_status("Missing MiqRequestTask id"))
     end
+
+    it "requires a provisioning object" do
+      floe_context = create_floe_context(FactoryBot.create(:service_reconfigure_task, :request_type => "service_reconfigure"))
+      runner_context = described_class.provision_execute(params, secrets, floe_context)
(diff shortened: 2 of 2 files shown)


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -20,6 +20,9 @@ def self.provision_execute(_params = {}, _secrets = {}, context = {})
           return BuiltinRunnner.error!({}, :cause => "Missing MiqRequestTask id")   if object_id.nil?
 
           miq_request_task = ::MiqRequestTask.find_by(:id => object_id.to_i)
+          new_options = miq_request_task.options.merge(context.input.symbolize_keys)
+          miq_request_task.update(:options => new_options)
```
Fryguy, line 24:
```suggestion
          miq_request_task.update!(:options => new_options)
```

Fryguy: Also these lines should come _after_ the 2 raise's below which check for the presence of the miq_request_task and the class type.

kbrock:
update:
- fixed `update!`
- add tests

update:
- cops

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -23,6 +23,8 @@ def self.provision_execute(_params = {}, _secrets = {}, context = {})
           return BuiltinRunnner.error!({}, :cause => "Unable to find MiqReqeustTask id: [#{object_id}]")                        if miq_request_task.nil?
           return BuiltinRunnner.error!({}, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless miq_request_task.class < ::MiqProvision
 
+          new_options = miq_request_task.options.merge(context.input.symbolize_keys)
+          miq_request_task.update!(:options => new_options)
```
agrare, line 27: @kbrock could you use `miq_request_task.options.merge!` then `miq_request_task.save!` to save creating a dup new object ?

kbrock: thnx.

tests still pass when `options_will_change!` was not used, but feels like it should be necessary.

kbrock:
update:
- using `merge!` and `save!`

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -23,6 +23,9 @@ def self.provision_execute(_params = {}, _secrets = {}, context = {})
           return BuiltinRunnner.error!({}, :cause => "Unable to find MiqReqeustTask id: [#{object_id}]")                        if miq_request_task.nil?
           return BuiltinRunnner.error!({}, :cause => "Calling provision_execute on non-provisioning request: [#{object_type}]") unless miq_request_task.class < ::MiqProvision
 
+          miq_request_task.options_will_change!
+          miq_request_task.options.merge!(context.input.symbolize_keys)
```
Fryguy, line 27: Not for this PR, but I wonder if we should only merge in keys that already exist?  That is, if the incoming input has extra stuff, we probably don't want to merge that in?  Not sure - would like to discuss first.

spec/lib/manageiq/providers/workflows/builtin_methods_spec.rb
```diff
@@ -17,14 +17,56 @@
...
+      expect(runner_context).to include("running" => false, "success" => false, "output" => failed_task_status("Missing MiqRequestTask id"))
     end
+
+    it "requires a provisioning object" do
+      floe_context = create_floe_context(FactoryBot.create(:service_reconfigure_task, :request_type => "service_reconfigure"))
+      runner_context = described_class.provision_execute(params, secrets, floe_context)
+      expect(runner_context).to include("running" => false, "success" => false, "output" => failed_task_status(/Calling provision_execute on non-provisioning request/))
+    end
+
+    it "requires a provisioning object" do
```
Fryguy, line 36: These 2 have the same title - not sure if that's a problem or if one clobbers the other.

spec/lib/manageiq/providers/workflows/builtin_methods_spec.rb
```diff
@@ -17,14 +17,56 @@
...
+    end
+  end
+
+  def create_floe_context(object = nil, execution: nil, input: {})
+    execution ||= {"_object_id" => object&.id, "_object_type" => object&.class}.compact
+
+    Floe::Workflow::Context.new({"Execution" => execution}, :input => input).tap { |ctx| ctx.state["Input"] = input }
+  end
+
+  def failed_task_status(cause = nil, error: "States.TaskFailed")
```
Fryguy, line 69: None of the callers actually pass error - is that intentional or was it missed?

kbrock: pretty much all errors will be `TaskFailed`.
So I was leaving this open, but probably not necessary.

You want me to simplify?

Fryguy: No, it's fine - I just wasn't sure if it was intentional or an oversight.

kbrock: holding off for now - may just drop in the future when adding more specs if it is never used

spec/lib/manageiq/providers/workflows/builtin_methods_spec.rb
kbrock: it was a problem. thanks

kbrock:
update:
- fixed specs with duplicate names
- only setting options that are known. (extra inputs will not be set)

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit d9166ff6b25c0953a60577892da0cf3140bb49f6.

```text
commit d9166ff6b25c0953a60577892da0cf3140bb49f6
Date:   Thu May 23 12:34:32 2024 -0400

    Merge pull request #94 from kbrock/sync-parameters
    
    automatically update options for provisioning requests
    
    (cherry picked from commit 973baba3a1d9dc31814e2b6118bcac5b9815d501)
```

PR #95, opened 2024-05-24
Fail seeding on invalid out of the box workflows.

Fryguy:
We should not ship invalid workflows, so raise an exception when seeding fails.

@agrare Please review.  Related to https://github.com/ManageIQ/manageiq/pull/23048, but doesn't really depend on it.


agrare (approved):
This works for now, if we move to marking workflows as invalid in the configuration_script_source we might need to rework this but can handle that then

PR #102, opened 2024-07-01
Add event driven workflows runner

agrare:
Add a Workflows::Runner class which keeps a persistent thread waiting for events from Floe.  Once an event is raised we try to link the event up to a workflow and we queue a step action if we find one.

If a workflow isn't found for example if it is already completed and removed from the list then we simply continue without queueing anything.

It is common to get multiple events for a single container run so this isn't a perfect "queue step when ready" and can be more efficient but currently this is erring on the side of caution (check step_ready on every event).

Also if the worker goes down and is restarted for any reason, the typical timer-based step is still on the queue and the running workflow will be added to the runner on the next timed check.

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23151
- [x] https://github.com/ManageIQ/floe/pull/265

Fixes https://github.com/ManageIQ/manageiq-providers-workflows/issues/109

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
@@ -49,13 +49,15 @@ class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < Mana
   end
 
   def run(args = {})
-    zone, role, object_type, object_id = args.values_at(:zone, :role, :object_type, :object_id)
+    queue_args = args.slice(:zone, :role, :object_type, :object_id)
+    zone, role, object_type, object_id = queue_args.values_at(:zone, :role, :object_type, :object_id)
+
+    ManageIQ::Providers::Workflows::Runner.runner.add_workflow(self, queue_args)
 
     object = object_type.constantize.find_by(:id => object_id) if object_type && object_id
     object.before_ae_starts({}) if object.present? && object.respond_to?(:before_ae_starts)
 
-    creds = resolved_credentials
-    wf = Floe::Workflow.new(payload, context, creds)
+    wf = Floe::Workflow.new(payload, context, resolved_credentials)
     wf.run_nonblock
     update_credentials!(wf.credentials)
 
@@ -76,7 +78,12 @@ class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < Mana
       object.after_ae_delivery(ae_result)
     end
 
-    run_queue(:zone => zone, :role => role, :object => object, :deliver_on => 10.seconds.from_now.utc, :server_guid => MiqServer.my_server.guid) unless wf.end?
+    if wf.end?
+      ManageIQ::Providers::Workflows::Runner.runner.delete_workflow(self)
+    else
+      deliver_on = wf.wait_until || 1.minute.from_now.utc
+      run_queue(:zone => zone, :role => role, :object => object, :deliver_on => deliver_on, :server_guid => MiqServer.my_server.guid)
+    end
   end
 
   private
--- a/lib/manageiq/providers/workflows/runner.rb
+++ b/lib/manageiq/providers/workflows/runner.rb
@@ -1,0 +1,86 @@
+module ManageIQ
+  module Providers
+    module Workflows
+      class Runner
+        include Vmdb::Logging
+
+        class << self
+          def runner
+            @runner ||= new.tap(&:start)
+          end
+        end
+
+        attr_reader :workflows
+
+        def initialize
+          require "floe"
+          require "concurrent/hash"
+
+          @workflows          = Concurrent::Hash.new
+          @docker_wait_thread = nil
+        end
+
+        def start
+          $workflows_log.debug("Runner: Starting workflows runner...")
+          self.docker_wait_thread = Thread.new { docker_wait }
+          $workflows_log.debug("Runner: Starting workflows runner...Complete")
+        end
+
+        def stop
+          $workflows_log.debug("Runner: Stopping workflows runner...")
+          stop_thread(docker_wait_thread)
+
+          self.docker_wait_thread = nil
+          $workflows_log.debug("Runner: Stopping workflows runner...Complete")
+        end
+
+        def add_workflow(workflow, queue_args)
+          workflows[workflow.id] ||= [workflow, queue_args]
+        end
+
+        def delete_workflow(workflow)
+          workflows.delete(workflow.id)
+        end
+
+        private
+
+        attr_accessor :docker_wait_thread
+
+        def docker_wait
+          loop do
+            docker_runner = Floe::Runner.for_resource("docker")
+            docker_runner.wait do |event, runner_context|
+              $workflows_log.info("Runner: Caught event [#{event}] for container [#{runner_context["container_ref"]}]")
+
+              workflow, queue_args = workflow_by_runner_context(runner_context)
+              next if workflow.nil?
+
+              $workflows_log.info("Runner: Queueing update for WorkflowInstance ID: [#{workflow.id}]")
+
+              workflow.run_queue(**queue_args)
+            end
+          rescue => err
+            $workflows_log.warn("Error: [#{err}]")
+            $workflows_log.log_backtrace(err)
+          end
+        end
+
+        def stop_thread(thread)
+          return if thread.nil?
+
+          thread.kill
+          thread.join(0)
+        end
+
+        def workflow_by_runner_context(runner_context)
+          workflows.detect do |_id, (workflow, _queue_args)|
+            context       = workflow.reload.context
+            container_ref = context.dig("State", "RunnerContext", "container_ref")
+
+            container_ref == runner_context["container_ref"]
+          end&.last
+        end
+      end
+    end
+  end
+end
--- a/lib/manageiq/providers/workflows/engine.rb
+++ b/lib/manageiq/providers/workflows/engine.rb
@@ -37,6 +37,10 @@ module ManageIQ
           ]
         end
 
+        def self.automation_runners
+          [ManageIQ::Providers::Workflows::Runner]
+        end
+
         def self.floe_runner_name
           if (runner_setting = Settings.ems.ems_workflows.runner.presence)
             runner_setting
--- a/manageiq-providers-workflows.gemspec
+++ b/manageiq-providers-workflows.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "floe", "~> 0.13.0"
+  spec.add_dependency "floe", "~> 0.13.0", ">= 0.13.1"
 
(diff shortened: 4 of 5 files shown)


lib/manageiq/providers/workflows/runner.rb
```diff
@@ -0,0 +1,91 @@
+module ManageIQ
+  module Providers
+    module Workflows
+      class Runner
+        include Vmdb::Logging
+
+        class << self
+          def runner
+            @runner ||= new.tap(&:start)
+          end
```
agrare, line 10: NOTE this ensures that this will work from the AutomationWorker as well as in development using simulate_queue_worker.

lib/manageiq/providers/workflows/runner.rb
```diff
@@ -0,0 +1,88 @@
...
+          stop_thread(docker_wait_thread)
+
+          self.docker_wait_thread = nil
+          $workflows_log.debug("Runner: Stopping workflows runner...Complete")
+        end
+
+        def add_workflow(workflow, queue_args)
+          return if workflows.key?(workflow.id)
+
+          workflows[workflow.id] = [workflow, queue_args]
```
kbrock, line 40: Does `workflows[] ||=` do the same thing with concurrent hashes?

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -49,13 +49,15 @@ def queue_callback(state, message, _result)
   end
 
   def run(args = {})
-    zone, role, object_type, object_id = args.values_at(:zone, :role, :object_type, :object_id)
+    queue_args = args.slice(:zone, :role, :object_type, :object_id)
+    zone, role, object_type, object_id = queue_args.values_at(:zone, :role, :object_type, :object_id)
```
kbrock, line 53: don't see where these are used

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -49,13 +49,15 @@ def queue_callback(state, message, _result)
...
+
+    ManageIQ::Providers::Workflows::Runner.runner.add_workflow(self, queue_args)
```
kbrock, line 55: note to self/please verify:
This is recorded so when an event comes from docker, it can be put back into the miq_queue.

- Are these zone/role values going to be the same for all messages?
- Is object_type/object_id pretty the workflow id?

agrare: > Are these zone/role values going to be the same for all messages?

Currently role will always be automate, I have these in case we wanted to do e.g. builtin workflows for ems_operations like we do with ansible_runner.

Zone can be different.

> Is object_type/object_id pretty the workflow id?

No object_type/object_id is typically a `ServiceTemplateProvisionTask` from a Service Order

agrare: These queue args are not the same for all workflow_instances which is why we are keeping the queue args with with specific workflow_instance.

lib/manageiq/providers/workflows/runner.rb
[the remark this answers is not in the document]
agrare: So aside from the boilerplate thread management, almost all of what is here is dealing with `ConfigurationScript` records and `MiqQueue`.  We could move more into Floe but we'd have to have callbacks for almost everything in the docker_wait method (currently it is basically just docker_runner.wait, find the `ConfigurationScript` record, miq_queue.put).

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
kbrock, line 53: nit: think you may be able to get away with keeping `s/queue_args/args/` (one less diff line)

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -76,7 +78,12 @@ def run(args = {})
       object.after_ae_delivery(ae_result)
     end
 
-    run_queue(:zone => zone, :role => role, :object => object, :deliver_on => 10.seconds.from_now.utc, :server_guid => MiqServer.my_server.guid) unless wf.end?
+    if wf.end?
+      ManageIQ::Providers::Workflows::Runner.runner.delete_workflow(self)
+    else
+      deliver_on = wf.wait_until || 1.minute.from_now.utc
```
kbrock, line 84: yay

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
agrare: I tried this originally but miq_queue automatically tacks on a miq_task_id to the args on delivery and that ended up throwing an exception because it wasn't a recognized keyword for run_queue

kbrock: Good catch. I really don't like that `$miq_task_id` logic

lib/manageiq/providers/workflows/runner.rb
```diff
@@ -0,0 +1,86 @@
...
+        end
+
+        private
+
+        attr_accessor :docker_wait_thread
+
+        def docker_wait
+          loop do
+            docker_runner = Floe::Runner.for_resource("docker")
+            docker_runner.wait do |event, runner_context|
```
kbrock, line 52: feeling like we only care about `:delete`

lib/manageiq/providers/workflows/runner.rb
```diff
@@ -0,0 +1,86 @@
...
+          return if thread.nil?
+
+          thread.kill
+          thread.join(0)
+        end
+
+        def workflow_by_runner_context(runner_context)
+          workflows.detect do |_id, (workflow, _queue_args)|
+            context       = workflow.reload.context
+            container_ref = context.dig("State", "RunnerContext", "container_ref")
```
kbrock, line 78: I'm concerned that we only support a single `container_ref`
feeling like we should register by the `container_ref`

But maybe when we need more (map) then we'll just update this.

If we register by container_ref, then we'll need a way to clear out old references (by workflow_id is nice)

agrare: Yes with Map/Parallel this will have to be tweaked to support that.

kbrock: If we do change this to container_ref, then this method becomes:

```
def workflow_by_runner_context(runner_context)
  workflow.delete(runner_context["container_ref"])
end
```

lib/manageiq/providers/workflows/runner.rb
agrare: Yes I agree this was "an abundance of caution" where we poke the queue more often that we have to, but without accidentally leaving a workflow "hung"

lib/manageiq/providers/workflows/runner.rb
```diff
@@ -0,0 +1,86 @@
...
+        def stop
+          $workflows_log.debug("Runner: Stopping workflows runner...")
+          stop_thread(docker_wait_thread)
+
+          self.docker_wait_thread = nil
+          $workflows_log.debug("Runner: Stopping workflows runner...Complete")
+        end
+
+        def add_workflow(workflow, queue_args)
+          workflows[workflow.id] ||= [workflow, queue_args]
```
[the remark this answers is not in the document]
kbrock: also, threw around the idea of passing a lambda instead of queue_args.

```ruby
-> { workflow.queue(queue_args] }
```

This will allow us to call with a Workflow instead of a `ScriptRunner`

kbrock:
The whole multiple events and keying off of workflow has me a little.

I feel the purpose of this is to react to an event from Docker
So every time you throw in a task, you should be waiting for the event to end and you will trigger your own event accordingly. It may be "put this on the Queue" or "put this on the "MiqQueue". And at that time, you delete the entry in the local reaction `workflows` cache, since we already reacted and have no need to react further.

If you run 3 tasks for a workflow, then you get 3 "I am done" events out and kick the same workflow accordingly.

I feel the entry should go into this cache when we kick off a task. Maybe this is where my idea of keying off of a `container_ref` breaks down. The container ref and the fact that we're calling a Task will be very well known to the caller. Feels like we're loosing some encapsulation.

lib/manageiq/providers/workflows/runner.rb
kbrock: yea. I worry about not coding the exact list of "done" and therefore not reacting because someone said "fire" in the wrong language

kbrock: If we have 10 parallel tasks, would be nice to know which Task to check.

lib/manageiq/providers/workflows/runner.rb
```diff
@@ -0,0 +1,86 @@
...
+        def delete_workflow(workflow)
+          workflows.delete(workflow.id)
+        end
+
+        private
+
+        attr_accessor :docker_wait_thread
+
+        def docker_wait
+          loop do
```
kbrock, line 50: does this loop handle an issue with docker

e.g.: in development we restart our docker process.
that will probably kick us out of `docker_runner.wait`

then the loop will swoop back, connect, and continue?

lib/manageiq/providers/workflows/runner.rb
agrare: I played with checking if it was `ready?` before putting the task on miq_queue which is essentially what step_nonblock_ready? will check anyway when we go to step maybe we go that direction (follow-up? this will work but less efficient)

lib/manageiq/providers/workflows/runner.rb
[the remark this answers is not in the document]
agrare: > I'm feeling like we should register by the container_ref

The container_ref changes constantly I don't think that we can index off of that, we'd be re-indexing every single lookup.

lib/manageiq/providers/workflows/runner.rb
agrare: I don't think we can delete from the workflow list on every event, also this assumes we are indexing on the container_ref which I don't think is practical since it can change every state in a workflow

lib/manageiq/providers/workflows/runner.rb
agrare: So both docker :tm: and podman use a command line utility to get events with a timeout, and kubernetes uses watches.  All three use a loop internally to retry on failure / break out if timeout has been exceeded.

kbrock:
offline: agrare doesn't want to store container_ref in the cache, it is too volatile and would prefer to store a workflow.id or something that is more stable

Fryguy:
Backported to `radjabov` in commit ad9bc0b2a3b19bb6c8ad9d5352efc3fe772a5235.

```text
commit ad9bc0b2a3b19bb6c8ad9d5352efc3fe772a5235
Date:   Tue Aug 20 10:27:18 2024 -0400

    Merge pull request #102 from agrare/add_event_driven_workflows_runner
    
    Add event driven workflows runner
    
    (cherry picked from commit 572047c0e74ce12269dffbac0520134804c8f1c4)
```

Issue #101, opened 2024-06-29
Expose Requester to workflows


kbrock:
In a workflow, if we want to send an email, we need to know the requester.

An example is when you provision a VM, you send an email to the requester to say that the vm has been provisioned.

Please make requester information (specifically email and name) available. Not sure if this needs to be accessible to the end user or if it is available to the email action.it just needs to be accessible to the email job

PR #105, opened 2024-07-17
Pass requester userid and email into workflows

kbrock:
I needed this for a demo

This allows us to send an email to the requester from a workflow

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -2,9 +2,10 @@ module ManageIQ
   module Providers
     module Workflows
       class BuiltinMethods < BasicObject
-        def self.email(params, _secrets, _context)
+        def self.email(params, _secrets, context)
           options = params.slice("To", "From", "Subject", "Cc", "Bcc", "Body", "Attachment").transform_keys { |k| k.downcase.to_sym }
           options[:from] ||= ::Settings.smtp.from
+          options[:to]   ||= context.execution["_requester_email"]
           miq_task = ::GenericMailer.deliver_task(:generic_notification, options)
 
           {"miq_task_id" => miq_task.id}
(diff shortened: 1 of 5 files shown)


kbrock:
Fixes #101

Fryguy:
@agrare Please review.

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -2,9 +2,10 @@ module ManageIQ
   module Providers
     module Workflows
       class BuiltinMethods < BasicObject
-        def self.email(params = {}, _secrets = {}, _context = {})
+        def self.email(params = {}, _secrets = {}, context = {})
           options = params.slice("To", "From", "Subject", "Cc", "Bcc", "Body", "Attachment").transform_keys { |k| k.downcase.to_sym }
           options[:from] ||= ::Settings.smtp.from
+          options[:to]   ||= context.execution["_requester_email"] if context.respond_to?(:execution)
```
agrare, line 8: When would context not respond to execution?  Do we actually ever pass in an actual hash or always a `Floe::Context` object?

kbrock: I think we sometimes pass the context in as a `Hash`.

agrare: Yeah if it can be a hash I think we should `context.dig("Execution", "_requester_email")` which should work with either, but it'd be better if we stay consistent with the types that we're passing in.

kbrock: we only ever pass in a Context, but the tests pass nothing, so they have a hash there.

Updated tests to pass a context (and match the way the methods are actually used)

spec/lib/manageiq/providers/workflows/builtin_runner_spec.rb
```diff
@@ -3,14 +3,16 @@
 
   let(:subject) { described_class.new(options) }
   let(:options) { {} }
+  let(:ctx)     { Floe::Workflow::Context.new }
 
   describe "run_async!" do
+
```
agrare, line 9:
```suggestion
```

agrare (approved):

PR #107, opened 2024-07-31
Bump floe to v0.12.0

agrare:
Floe switched to enforcing workflow input and output is JSON so we need to update our callers here


Fryguy:
Backported to radjabov via merge of master into radjabov branch

PR #112, opened 2024-08-20
Use workflow_instance manager_ref as Execution ID

agrare:
Set the WorkflowInstance manager_ref as the Context.Execution.Id so that we can use it to find the WorkflowInstance more easily

Related:
- [ ] https://github.com/ManageIQ/floe/pull/268

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow.rb
@@ -10,12 +10,18 @@ class ManageIQ::Providers::Workflows::AutomationManager::Workflow < ManageIQ::Pr
   def run(inputs: {}, userid: "system", zone: nil, role: "automate", object: nil, execution_context: {})
     require "floe"
 
+    manager_ref = SecureRandom.uuid
+
     execution_context = execution_context.dup
+
+    execution_context["Id"]                = manager_ref
     execution_context["_manageiq_api_url"] = MiqRegion.my_region.remote_ws_url
+
     if object
       execution_context["_object_type"] = object.class.name
       execution_context["_object_id"]   = object.id
     end
+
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/workflows/automation_manager/workflow.rb
```diff
@@ -10,12 +10,18 @@ def self.create_from_json!(json, **kwargs)
   def run(inputs: {}, userid: "system", zone: nil, role: "automate", object: nil, execution_context: {})
     require "floe"
 
+    manager_ref = SecureRandom.uuid
+
     execution_context = execution_context.dup
+
+    execution_context["Id"]                = manager_ref
```
kbrock, line 17: do we want the same name in both places for consistency?
Also, I thought you tended to like an underscore before these entries

agrare: Execution Id is one of the predefined values in the Context object https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html so I thought it would make sense to map that to our equivalent of a StateMachine Execution (WorkflowInstance)

kbrock: reading the floe code - I really like execution id

Fryguy:
Backported to `radjabov` in commit 416dd919938a893a31d5f0716313a27f87114727.

```text
commit 416dd919938a893a31d5f0716313a27f87114727
Date:   Tue Aug 20 12:26:24 2024 -0400

    Merge pull request #112 from agrare/set_execution_id_as_manager_ref
    
    Use workflow_instance manager_ref as Execution ID
    
    (cherry picked from commit 2f6a6ad5f7e64b7e9944e19b518777ed77361899)
```

PR #113, opened 2024-08-20
Use WorkflowInstance#manager_ref and execution_id for events

agrare:
Use the execution_id from the docker event to lookup the workflow_instance rather than the container_ref.  Since this is static for the life of the workflow it is a much more reliable way to make the association.

Depends on:
- [x] https://github.com/ManageIQ/floe/pull/268
- [x] Release of https://github.com/ManageIQ/floe/pull/268

### DIFF
--- a/lib/manageiq/providers/workflows/runner.rb
+++ b/lib/manageiq/providers/workflows/runner.rb
@@ -10,13 +10,13 @@ module ManageIQ
           end
         end
 
-        attr_reader :workflows
+        attr_reader :workflow_instances
 
         def initialize
           require "floe"
           require "concurrent/hash"
 
-          @workflows          = Concurrent::Hash.new
+          @workflow_instances = Concurrent::Hash.new
           @docker_wait_thread = nil
         end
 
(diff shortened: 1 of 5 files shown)


lib/manageiq/providers/workflows/runner.rb
```diff
@@ -34,12 +34,12 @@ def stop
           $workflows_log.debug("Runner: Stopping workflows runner...Complete")
         end
 
-        def add_workflow(workflow, queue_args)
-          workflows[workflow.id] ||= [workflow, queue_args]
+        def add_workflow_instance(workflow_instance, queue_args)
+          workflow_instances[workflow_instance.manager_ref] ||= [workflow_instance, queue_args]
```
Fryguy, line 38: Not sure if this matters but does `||=` work the way you expect with a concurrent Hash?  (or does it matter here)

I thought `x ||= y` was syntactic sugar for `x || x = y`, meaning this would be a check and then a set and would be non-atomic.

Fryguy:
Backported to `radjabov` in commit e7dd272c0069375c400f0c6ea234dd95405c2b9b.

```text
commit e7dd272c0069375c400f0c6ea234dd95405c2b9b
Date:   Wed Aug 21 11:34:46 2024 -0400

    Merge pull request #113 from agrare/use_docker_event_execution_id
    
    Use WorkflowInstance#manager_ref and execution_id for events
    
    (cherry picked from commit b6fbd9a0fc8d20cfef740bec2a92e6809045ae49)
```

PR #121, opened 2024-11-18
Add object_type,_id to WorkflowInstance#run_queue

agrare:
Fix an issue where the `Workflows::Runner#docker_wait` thread queues the `workflow_instance` with `queue_args` that have `object_type`/`object_id` and not the resolved object.

`WorkflowInstance#run` looks the object up by type,id so it has the object already but `Workflows::Runner#docker_wait` only has the raw `queue_args` so it is better to allow `object_type`,`object_id` in
`WorkflowInstance#run_queue`.

### DIFF
--- a/lib/manageiq/providers/workflows/runner.rb
+++ b/lib/manageiq/providers/workflows/runner.rb
@@ -22,7 +22,7 @@ module ManageIQ
 
         def start
           $workflows_log.debug("Runner: Starting workflows runner...")
-          self.docker_wait_thread = Thread.new { docker_wait }
+          self.docker_wait_thread = Thread.new { loop { docker_wait } }
           $workflows_log.debug("Runner: Starting workflows runner...Complete")
         end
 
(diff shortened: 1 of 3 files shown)


agrare:
I'm going to try to add a spec that would have caught this

spec/lib/manageiq/providers/workflows/runner_spec.rb
```diff
@@ -29,4 +29,58 @@
...
+          subject.send(:docker_wait)
+
+          expect(MiqQueue.count).to eq(1)
+          expect(MiqQueue.first).to have_attributes(
+            :class_name  => "ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance",
+            :method_name => "run",
+            :args        => [{:zone => nil, :role => "automate", :object_type => service.class.name, :object_id => service.id}],
+          )
+        end
+      end
```
agrare, line 83: This was the case that caused the original bug

lib/manageiq/providers/workflows/runner.rb
```diff
@@ -47,23 +47,21 @@ def delete_workflow_instance(workflow_instance)
         attr_accessor :docker_wait_thread
 
         def docker_wait
-          loop do
```
agrare, line 50: I moved the loop up a level so it was easier to test this method in isolation, think it actually makes more sense since the loop is just for the thread to continue if the wait terminates prematurely for some reason.

agrare:
Okay @Fryguy added specs covering the docker_wait method in the Runner

lib/manageiq/providers/workflows/runner.rb
Fryguy: What breaks out of the loop though?

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit 71307dd2345e7abe94040aa39adb03058b1285c4.

```text
commit 71307dd2345e7abe94040aa39adb03058b1285c4
Date:   Tue Nov 19 10:00:25 2024 -0500

    Merge pull request #121 from agrare/fix_object_type_id_argument_error
    
    Add object_type,_id to WorkflowInstance#run_queue
    
    (cherry picked from commit 8cacfab2e740546a5f529b82ddbbca50829cf618)
```

PR #124, opened 2025-01-14
Skip invalid workflows during sync

agrare:
When syncing a repository and creating workflows check if the payload is valid and if it isn't log a (hopefully) helpful error message.

This is based on https://github.com/ManageIQ/manageiq-providers-workflows/pull/59 but instead of requiring a new column this will log the error and skip any invalid workflows.

There was also an issue with the not-implemented states raising an exception that wasn't a `Floe::InvalidWorkflowError` which would have skipped this rescue.

Example log message:
`WARN -- evm: MIQ(ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSource#create_workflow_from_payload) Invalid ASL file [missing_states.asl]: State Machine does not have required field "States"`

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
@@ -55,7 +55,8 @@ class ManageIQ::Providers::Workflows::AutomationManager::ConfigurationScriptSour
 
         payload  = worktree.read_file(filename)
         workflow = create_workflow_from_payload(filename, payload, fail_on_invalid_workflow: false)
-        to_delete.delete(workflow.name)
+
+        to_delete.delete(workflow.name) if workflow
       end
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/workflows/automation_manager/configuration_script_source.rb
```diff
@@ -68,31 +69,34 @@ def sync_from_content(to_delete)
...
     end
   end
 
   def create_workflow_from_payload(name, payload, fail_on_invalid_workflow:)
-    floe_workflow =
+    floe_workflow, payload_error =
       begin
         Floe::Workflow.new(payload)
-      rescue Floe::InvalidWorkflowError
+      rescue Floe::InvalidWorkflowError, NotImplementedError => err
```
agrare, line 81: NOTE there was an issue in floe prior to version v0.15.0 where Map and Parallel raised a `NotImplementedError` rather than a `Floe::InvalidWorkflowError`

Fryguy (approved):

PR #128, opened 2025-04-01
Set logger for each workflow run

kbrock:
Blocked by:
- [x] https://github.com/ManageIQ/manageiq/pull/23426

Creates RequestLog records for each log entries
This way the messages will show up in The log entries on the Request screen This creates the same.

Did feel like I reinvented the wheel with the log proxy, but it seemed the simplest way.
I lifted `request_log` - with only trivial changes

SCR-20250401-qjhv

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
@@ -60,7 +60,9 @@ class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < Mana
     object = object_type.constantize.find_by(:id => object_id) if object_type && object_id
     object.before_ae_starts({}) if object.present? && object.respond_to?(:before_ae_starts)
 
-    wf = Floe::Workflow.new(payload, context, resolved_credentials)
+    context_obj = Floe::Workflow::Context.new(context, :logger => create_logger(object))
+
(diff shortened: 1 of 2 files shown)


Fryguy:
Trying to understand here, but is this not the same as the request logger from the automation engine? https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/manageiq/automation_engine/logger.rb

Should we just use that one?

kbrock:
update:
- now using logger from core

app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
```diff
@@ -60,7 +60,9 @@ def run(args = {})
     object = object_type.constantize.find_by(:id => object_id) if object_type && object_id
     object.before_ae_starts({}) if object.present? && object.respond_to?(:before_ae_starts)
 
-    wf = Floe::Workflow.new(payload, context, resolved_credentials)
+    context_obj = Floe::Workflow::Context.new(context, logger: create_logger(object))
```
agrare, line 63:
```suggestion
    context_obj = Floe::Workflow::Context.new(context, :logger => create_logger(object))
```

kbrock:
update:
- fixed hash rocket syntax
- fixed test to make more lenient around context (we are testing passwords not context)

agrare (approved):

Fryguy:
Backported to spassky via merge of master into spassky

PR #130, opened 2025-05-13
Introduce embedded_terraform builtin

kbrock:
This allows users to run a workflow for embedded terraform

```
    "PreProcess": {
      "Type": "Task",
      "Resource": "manageiq://embedded_terraform",
      "Parameters": {
        "Stage": "preprocess"
      },
      "Next": "Execute"
    },
    "Execute": {
      "Type": "Task",
      "Resource": "manageiq://embedded_terraform",
      "Parameters": {
        "Stage": "execute"
      },
      "Next": "Refresh"
    },
    "Refresh": {
      "Type": "Task",
      "Resource": "manageiq://embedded_terraform",
      "Parameters": {
        "Stage": "refresh"
      },
      "Next": "PostProcess"
    },
    "PostProcess": {
      "Type": "Task",
      "Resource": "manageiq://embedded_terraform",
      "Parameters": {
        "Stage": "postprocess"
      },
      "Next": "Email"
    },
```

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -74,6 +74,78 @@ module ManageIQ
           miq_request_task_status!(runner_context)
         end
 
+        def self.embedded_terraform(params, _secrets, context)
+          object_type, object_id = context.execution.values_at("_object_type", "_object_id")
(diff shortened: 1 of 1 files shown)


Fryguy:
As discussed I think this would be preferable to be moved to an internal StateMachine/Job similar to what we do for other things, and ideally the called would just do a single `manageiq://provision_execute`.

PR #131, opened 2025-06-03
Change podman root containers/storage directory

agrare:
https://github.com/ManageIQ/manageiq/issues/23470


kbrock:
I would like to not assume production is on an appliance.
Maybe we could use an env variable (i.e.: `TMP_DIR`?) or something to do this instead?

agrare:
Yeah we could probably check `MiqEnvironment.is_appliance?` instead

PR #132, opened 2025-07-01
Switch to use VcrSecrets helper from core

Fryguy:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23497


Fryguy:
@agrare We don't seem to use secrets here, so should we drop it?

agrare (approved):

Issue #134, opened 2025-07-16
Workflow run fails when using a RequestLogger


agrare:
```
[NoMethodError]: undefined method `info' for an instance of Array  Method:[block in method_missing]
  lib/floe/workflow/state.rb:73:in `mark_started'
  lib/floe/workflow/state.rb:55:in `start'
  lib/floe/workflow/state.rb:46:in `run_nonblock!'
  lib/floe/workflow.rb:111:in `step_nonblock'
  lib/floe/workflow.rb:103:in `run_nonblock'
  app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb:66:in `run'
```

PR #135, opened 2025-07-16
Fix Logger error when called from service

agrare:
When called from a MiqRequest object we use a `Vmdb::Loggers::RequestLogger` so that floe logs can be seen in the UI easily.  If `Floe.logger` is already a `BroadcastLogger` then `Floe.logger.wrap` isn't calling `ManageIQ::Loggers::Base#wrap` it is actually calling `ActiveSupport::BroadcastLogger#method_missing` which does `loggers.map { |logger| logger.send(name, ...) }` and results in an array being returned.

This array is then used as the `workflow.context.logger` so when we try to log to it we get `undefined method info for Array`

Introduced by https://github.com/ManageIQ/manageiq-providers-workflows/pull/128
Fixes https://github.com/ManageIQ/manageiq-providers-workflows/issues/134

### DIFF
--- a/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
+++ b/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb
@@ -137,6 +137,9 @@ class ManageIQ::Providers::Workflows::AutomationManager::WorkflowInstance < Mana
 
   def create_logger(object)
     log_target_id = object.miq_request_id if object.respond_to?(:miq_request_id)
-    log_target_id ? Floe.logger.wrap(Vmdb::Loggers::RequestLogger.new(:resource_id => log_target_id)) : Floe.logger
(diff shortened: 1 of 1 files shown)


agrare:
NOTE we need to review usage of the instance method `#wrap` and ensure it can't be called on an object which is already a `BroadcastLogger`.

bdunne:
Backported to spassky in 1c071ac9eded154f1376285de55d6df2cef3a510
```
commit 1c071ac9eded154f1376285de55d6df2cef3a510
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Wed Jul 16 08:03:41 2025 -0700

    Merge pull request #135 from agrare/fix_broadcast_logger_error
    
    Fix Logger error when called from service
    
    (cherry picked from commit 5d28b7af49ad0fb4fc0ebe4dad73331431001c6e)
```

PR #138, opened 2025-08-11
Fix Typo BuiltinRunnner

agrare:
This looks like it was a typo back to the introduction of the class in 70d79b57 but was consistent so it was working.


Fryguy:
Is this technically a breaking change?

kbrock:
Thank you. This was annoying

PR #139, opened 2025-08-12
Use workflow credentials interface

kbrock:
depends:
- add interface to floe and release floe

This removes the knowledge of the context constructor and the workflow constructor It also standardizes on passing around ruby hashes and not json strings


kbrock:
update:
- go from large changes to just changing credentials

update:
- fix rspec - it is very implementation dependent, but works great

@agrare do we want to hold off on this change until after we merge https://github.com/ManageIQ/floe/pull/308 ?

agrare (approved):

PR #144, opened 2025-08-25
Add a manageiq://api builtin method

agrare:
Add a builtin method specific to the ManageIQ API.  This uses the `Context.Execution._manageiq_api_url` which is populated by the Embedded Workflows provider when executing a Workflow, based on `MiqRegion.my_region.remote_ws_url`: https://github.com/ManageIQ/manageiq-providers-workflows/blob/master/app/models/manageiq/providers/workflows/automation_manager/workflow.rb#L18C46-L18C79

It also defaults to JSON encoding and application/json ContentType.

We could probably do something with authentication as well.

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -2,6 +2,25 @@ module ManageIQ
   module Providers
     module Workflows
       class BuiltinMethods < Floe::BuiltinRunner::Methods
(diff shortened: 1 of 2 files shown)


agrare:
> It does get me thinking about pagination. Perhaps as a follow up or another Option of some sort.

Yeah we could do something with the "async" mode where we request a page at a time without blocking in the initial call

Fryguy:
Do we need docs?

PR #146, opened 2025-09-04
Prevent Floe wait thread from starting in tests

agrare:
`WorkflowInstance#run` calls into `ManageIQ::Providers::Workflows::Runner` to add/remove the workflow instance, and on initial access to the runner instance the wait event thread is started.


spec/models/manageiq/providers/workflows/automation_manager/workflow_instance_spec.rb
```diff
@@ -21,6 +21,15 @@
     }
   end
 
+  before do
+    runner_mock = double("ManageIQ::Providers::Workflows::Runner")
+
+    allow(runner_mock).to receive(:add_workflow_instance).and_return(nil)
+    allow(runner_mock).to receive(:delete_workflow_instance).and_return(nil)
```
kbrock, line 28: do you have a preference for this vs

```suggestion
    runner_mock = double("ManageIQ::Providers::Workflows::Runner", :add_workflow_instance => nil, :delete_workflow_instance => nil)
```

kbrock (approved):
LGTM
only 1 question

spec/models/manageiq/providers/workflows/automation_manager/workflow_instance_spec.rb
agrare: :+1: yeah that works, always forget `double()` can do that.  That part of my brain is taken up with LVM knowlege :rofl:

PR #157, opened 2026-06-12
Include info about the MiqRequestTask in state output

agrare:
Include the ID, href, state, status, and source/destination information in the state output.

This will help further states by including what is needed directly or by retrieving additional info from the API.

Example from a `manageiq://provision_execute` state:
```
    {"id"=>122,
     "href"=>"http://localhost:3000/api/request_tasks/122",
     "state"=>"provisioned",
     "source"=>{"id"=>800, "href"=>"http://localhost:3000/api/templates/800"},
     "status"=>"Ok",
     "destination"=>{"id"=>800, "href"=>"http://localhost:3000/api/vms/51"}}
```

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -9,7 +9,7 @@ module ManageIQ
             path         = params.delete("Path") || "/"
             api_base_url = context.execution["_manageiq_api_url"]
 
-            params["Url"] ||= ::URI.join(api_base_url, path).to_s
+            params["Url"] ||= ::File.join(api_base_url, path).to_s
           end
 
           params["Options"] ||= {}
@@ -92,7 +92,7 @@ module ManageIQ
           miq_request_task.save!
           miq_request_task.execute_queue
 
-          {"miq_request_task_id" => miq_request_task.id}
+          {"miq_request_task_id" => miq_request_task.id, "_manageiq_api_url" => context.execution&.dig("_manageiq_api_url")}
         end
 
         private_class_method def self.provision_execute_status!(runner_context)
@@ -115,7 +115,7 @@ module ManageIQ
           miq_request_task.save!
           miq_request_task.execute_queue
 
-          {"miq_request_task_id" => miq_request_task.id}
+          {"miq_request_task_id" => miq_request_task.id, "_manageiq_api_url" => context.execution&.dig("_manageiq_api_url")}
         end
 
         private_class_method def self.retire_execute_status!(runner_context)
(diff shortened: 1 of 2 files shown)


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -150,9 +150,18 @@ def self.retire_execute(params, _secrets, context)
             runner_context["running"] = true
             runner_context
           when "ok"
-            BuiltinRunner.success!(runner_context, :output => {"Result" => miq_request_task.completed_state})
+            BuiltinRunner.success!(runner_context, :output => miq_request_task_result(miq_request_task))
           end
         end
+
+        private_class_method def self.miq_request_task_result(miq_request_task)
+          api_base_url = "http://localhost:3000/api"
```
agrare: TODO we get the `_manageiq_api_url` in state context which isn't immediately available here

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -150,9 +150,28 @@ def self.retire_execute(params, _secrets, context)
             runner_context["running"] = true
             runner_context
           when "ok"
-            BuiltinRunner.success!(runner_context, :output => {"Result" => miq_request_task.completed_state})
+            BuiltinRunner.success!(runner_context, :output => miq_request_task_result(runner_context, miq_request_task))
           end
         end
+
+        private_class_method def self.miq_request_task_result(runner_context, miq_request_task)
+          api_base_url = ::File.join(runner_context["_manageiq_api_url"], "api") if runner_context["_manageiq_api_url"]
```
Fryguy, line 164: Oh weird - I thought the _manageiq_api_url _already_ had the /api bit on it.

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -9,7 +9,7 @@ def self.api(params, secrets, context)
             path         = params.delete("Path") || "/"
             api_base_url = context.execution["_manageiq_api_url"]
```
Fryguy, line 10: We don't add `/api` here, so is the Path expected to have /api in it?

lib/manageiq/providers/workflows/builtin_methods.rb
agrare: This comes from `MiqRegion.my_region.remote_ws_url` which for me is just:
```
vmdb(dev)> MiqRegion.my_region.remote_ws_url
=> "http://localhost:4000"
```

https://github.com/ManageIQ/manageiq-providers-workflows/blob/master/app/models/manageiq/providers/workflows/automation_manager/workflow.rb#L18

lib/manageiq/providers/workflows/builtin_methods.rb
agrare: Yes, `floe://http` expects full path relative to the base url and this doesn't prepend anything.

Fryguy (approved):

PR #159, opened 2026-06-29
Add Username/Password options to `manageiq://api`

agrare:
Not sure we want to do this, but I had this locally and it simplified the step of having to login and set the token in credentials before using the API and figured I'd throw it up for comment

### DIFF
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -18,6 +18,10 @@ module ManageIQ
           params["Headers"] ||= {}
           params["Headers"]["ContentType"] ||= "application/json"
 
+          if secrets["username"] && secrets["password"]
+            params["Headers"]["Authorization"] = "Basic #{::Base64.strict_encode64(secrets.values_at("username", "password").join(":"))}"
+          end
+
           http(params, secrets, context)
         end
 
--- a/lib/manageiq/providers/workflows/builtin_methods.rb
+++ b/lib/manageiq/providers/workflows/builtin_methods.rb
@@ -18,7 +18,9 @@ module ManageIQ
           params["Headers"] ||= {}
           params["Headers"]["ContentType"] ||= "application/json"
 
-          if secrets["username"] && secrets["password"]
+          if secrets["bearer_token"]
+            params["Headers"]["Authorization"] = "Bearer #{secrets["bearer_token"]}"
+          elsif secrets["username"] && secrets["password"]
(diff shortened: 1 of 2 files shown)


lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -18,6 +18,10 @@ def self.api(params, secrets, context)
           params["Headers"] ||= {}
           params["Headers"]["ContentType"] ||= "application/json"
 
+          if secrets["username"] && secrets["password"]
+            params["Headers"]["Authorization"] = "Basic #{::Base64.urlsafe_encode64("#{secrets["username"]}:#{secrets["password"]}")}"
```
Fryguy: >  urlsafe_encode64 replaces + and / with - and _. If your credentials happen to generate these characters, a standard Basic Auth parser on the server will fail to decode them. Even though it is inside an HTTP header, header values do not use URL-safe encoding.

```suggestion
            params["Headers"]["Authorization"] = "Basic #{::Base64.strict_encode64("#{secrets["username"]}:#{secrets["password"]}")}"
```

Fryguy: Also maybe a little cleaner? (I could give or take this one)

```suggestion
            params["Headers"]["Authorization"] = "Basic #{::Base64.urlsafe_encode64("#{secrets.values_at("username", "password").join(":"))}"
```

agrare: :+1: done

agrare: :+1: done

lib/manageiq/providers/workflows/builtin_methods.rb
```diff
@@ -18,6 +18,10 @@ def self.api(params, secrets, context)
           params["Headers"] ||= {}
           params["Headers"]["ContentType"] ||= "application/json"
 
+          if secrets["username"] && secrets["password"]
+            params["Headers"]["Authorization"] = "Basic #{::Base64.urlsafe_encode64(secrets.values_at("username", "password").join(":"))}"
```
Fryguy: oops the strict change got lost.


```suggestion
            params["Headers"]["Authorization"] = "Basic #{::Base64.strict_encode64(secrets.values_at("username", "password").join(":"))}"
```

Fryguy (approved):

Fryguy (approved):

Fryguy (approved):

